PR #30, opened 2014-12-11
remove mysql_config's options not compatible with MariaDB

bigio:
Some options of the program mysql_config are not compatible with MariaDB (they have been removed), use only the options compatible both with MySQL and MariaDB

### DIFF
--- a/Makefile.PL
+++ b/Makefile.PL
@@ -19,23 +19,9 @@ our $opt = { "help" => \&Usage, };
 Getopt::Long::GetOptions(
     $opt,
     "help",
-    "testdb=s",
-    "testhost=s",
-    "testport=s",
-    "testuser=s",
-    "testpassword=s",
-    "testsocket=s",
     "cflags=s",
     "libs=s",
-    "verbose",
-    "ps-protocol",
-    "bind-type-guessing",
-    "nocatchstderr",
-    "ssl!",
-    "nofoundrows!",
-    "embedded=s",
     "mysql_config=s",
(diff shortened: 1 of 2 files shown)


mbeijen:
I understand that problem - I addressed it now in a8f66357a67c0f1595854eed877fa7a1b4a2b609 . Your suggested solution would remove command line options for the Makefile.PL - I leave that still in place while also only using mysql_config for options such as --libs.

Please let me know if this works for you, or if you still see the verbose output.

bigio:
On 01/28/15 21:03, Michiel Beijen wrote:
> I understand that problem - I addressed it now in a8f6635 <https://github.com/perl5-dbi/DBD-mysql/commit/a8f66357a67c0f1595854eed877fa7a1b4a2b609> . Your suggested solution would remove command line options for the Makefile.PL - I leave that still in place while also only using mysql_config for options such as --libs.
> 
> Please let me know if this works for you, or if you still see the verbose output.
> 
it works, thanks.
 Cheers
  Giovanni

dveeden:
It looks like this PR should be closed?

bigio:
On 05/14/15 19:45, Daniël van Eeden wrote:
> It looks like this PR should be closed?
> 
> —
> Reply to this email directly or view it on GitHub <https://github.com/perl5-dbi/DBD-mysql/pull/30#issuecomment-102114982>.
> 
Yes, it should be closed.
 Giovanni

mbeijen:
Closing!

PR #31, opened 2014-12-23
Prepared statement and mysql v5.6+ fixes

stevenh:
This includes the following fixes, as per individual commits:
* Server prepared statement not dealing with truncated response from mysql_stmt_fetch
* Prepared statement version corrections.
* Support testing innodb under mysql 5.6+
* undef warning in DbiError if triggered.

This passes a full test run no FreeBSD 10.1 against mysql 5.6.22 including extended tests.


mbeijen:
Merged into master! thanks a lot.

PR #32, opened 2015-01-27
Fix for Windows and bug 68374 (bit type)

CaptTofu:
First pass - all tests good.


stevenh:
The max change here is incorrect, will do a pull request shortly.

stevenh:
Fix for build under windows is here: https://github.com/perl5-dbi/DBD-mysql/pull/33

PR #34, opened 2015-01-29
Allow several spaces after LIMIT clause.

dwest1975:
Previously the need to interpret bind variable as number was determined
based on presence of 'limit ?', 'Limit ?', 'lIMIT ?', 'LIMIT ?' in the
query. I.e. if there is indeed a LIMIT clause and it is followed by a
single whitespace and a bind variable placeholder. This makes it
impossible to write queries with multiple whitespaces after LIMIT clause
without explicit type binding.

### DIFF
--- a/dbdimp.c
+++ b/dbdimp.c
@@ -607,8 +607,8 @@ static char *parse_params(
         it would be good to be able to handle any number of cases and orders
       */
       if ((*statement_ptr == 'l' || *statement_ptr == 'L') &&
-          (!strncmp(statement_ptr+1, "imit ?", 6) ||
-           !strncmp(statement_ptr+1, "IMIT ?", 6)))
+          (!strncmp(statement_ptr+1, "imit ", 5) ||
+           !strncmp(statement_ptr+1, "IMIT ", 5)))
       {
         limit_flag = 1;
       }


mbeijen:
Hi, thanks for your fix.

However I think this is an issue that needs a better fix. For instance, the code does weird case-checking; even with the current fix you can't use `... limIT ?`. But that's just on the surface: also using `LIMIT ? OFFSET ?` does not work. 

http://dev.mysql.com/doc/refman/5.6/en/select.html

Would you be able with a fix that makes this possible? Basically, ? should be parsed as a placeholder when it's not in a comment or a quote, there should be no special cases for 'LIMIT' or other keywords.

If you'd not be able (time/skillwise) to do so, that's perfectly OK, just let me know!

dwest1975:
Hi,

Hm, I get the case-checking part of the requirement. That's easy to fix/improve.

But I'm not sure about `LIMIT ? OFFSET ?`. Currently it works like a charm (tested on 4.023, server - 5.6.20). Although with the same limitations as bare `LIMIT` has.

Also if a special case for `LIMIT` is to be removed it will break backward compatibility. I.e. I suppose many users do not explicitly bind variable types for limit/offset and DBD::mysql will quote these variables by default making incorrect SQL syntax. Just to illustrate:

```
SELECT * FROM Centrex_Queues LIMIT  ? OFFSET ?" with ParamValues: 0=1, 1=0]
DBD::mysql::st execute failed: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ''1' OFFSET '0'' at line 2 [for Statement "/* PID: 13766; Script: .e; */
```

pali:
Case insensitive check can be done e.g. by strncasecmp() which is in POSIX.

CaptTofu:
I wrote the original code around emulated prepare with regard to LIMIT, so I'll look at this PR so the fix can be progressed

pali:
Slightly different change with support for upper/lower case LIMIT, including unit tests for verification I included into DBD::MariaDB.

PR #35, opened 2015-03-09
Determine order for the selection to ensure the correct row is returned

Scimon:
We've found this test is failing during our build process as the SELECT is returning the row with id_binary_test == 2.
I wasn't sure if this was intentional though from reading the bug request it didn't seem to be. By adding an ORDER BY clause the id == 1 row is always returned first allowing the test to perform normally.

### DIFF
--- a/t/rt88006-bit-prepare.t
+++ b/t/rt88006-bit-prepare.t
@@ -33,7 +33,7 @@ ok $dbh->do($create),"create table for $scenario";
 
 ok $dbh->do("INSERT INTO `dbd_mysql_rt88006_bit_prep` (`id_binary_test`, `flags`) VALUES (2, b'10'), (1, b'1')");
 
-my $sth = $dbh->prepare("SELECT id_binary_test,flags FROM dbd_mysql_rt88006_bit_prep");
+my $sth = $dbh->prepare("SELECT id_binary_test,flags FROM dbd_mysql_rt88006_bit_prep ORDER BY id_binary_test ASC");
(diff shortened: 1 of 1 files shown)


CaptTofu:
Thank you for informing us of this!

Scimon:
Not a problem. :)

mbeijen:
Does your build process run against the 5.0 release of MySQL Server? I found these CPAN Testers results: http://www.cpantesters.org/cpan/report/e66c657a-c1ac-11e4-bab0-56037db652c0

I set up mysql server 5.0 on my local machine and was able to reproduce these results. And after applying your patch tests pass.

Scimon:
It's 5.1.73-0ubuntu0.10.04.1 on the server and Ver 14.14 Distrib 5.1.73 on the client. (Updating the the 14.04 LTS is on the roadmap) mysql_config is also showing 5.1.73.

mbeijen:
Thanks; it's not per se the server version - it's the storage engine. If it's InnoDB, the test passes, on MyISAM, the test fails. But the out-of-the-box default on 5.1 and up is InnoDB, and on 5.0 and below is MyISAM. This is why the smoker tests fail on 5.0.

I'll add an indicator to the test suite to show the default storage engine.

Scimon:
Ahhh. So our (very old) build setup may be defaulting to MyISAM. (I know we specify table engines when creating tables).
I'll check that out.

mbeijen:
I improved the test a bit here: https://github.com/perl5-dbi/DBD-mysql/commit/6fe0a1db08c12945ccc16e3804bb9aaf93a0cc77

PR #37, opened 2015-05-14
Add support for utf8mb4

dveeden:
This is to use utf8mb4 from the start of the connection. This is to allow 4-byte UTF-8 characters.


mbeijen:
Thanks! I really appreciated that you also took care to update docs & tests.

--
Michiel

PR #40, opened 2015-07-24
fix the check of InnoDB support

bor:
make sure what InnoDB engine is not just supported but enabled also

### DIFF
--- a/t/40catalog.t
+++ b/t/40catalog.t
@@ -47,7 +47,7 @@ SKIP: {
     } else {
        foreach my $engine (@$engines) {
          if (lc($engine->[0]) eq 'innodb') {
-           $have_innodb = 'YES';
+           $have_innodb = $engine->[1];
          }
        }
     }


bor:
I have disabled InnoDB engine in config.
And I've got (without the fix):

``
t/40catalog.t .. 1/78 DBD::mysql::db do failed: Unknown storage engine 'INNODB' at t/40catalog.t line 60.
DBD::mysql::db do failed: Unknown storage engine 'INNODB' at t/40catalog.t line 60.
``

mbeijen:
Thanks for raising the pull request.

The documentation says:

> A value of YES, NO, or DEFAULT indicates that an engine is available, not available, or available and currently set as the default storage engine. 

http://dev.mysql.com/doc/refman/5.6/en/storage-engines.html

With your fix it only actually runs the test if the value would be `YES` and not with `DEFAULT`. I modified the test so it skips the engine if it is installed but set to `NO`. See the commit below:

https://github.com/perl5-dbi/DBD-mysql/commit/139c3783b438e470f6e9d2f86496c9200eb53f49

Kindest regards,

Michiel

PR #41, opened 2015-07-24
fix 'create table' query to make it suitable for any kind of engine

bor:
do not set an engine directly but leave it as default
personally I've tested it on MyISAM and all tests passed


bor:
Another way is just skip such tests if InnoDB is not supported.
But I think tests should works well for any kind of engines.
What do you think?

CaptTofu:
I will look at this

CaptTofu:
and thank you!

PR #42, opened 2015-09-28
Use documented API to get net_buffer_length

berntm:
net_buffer_length and mysql_get_parameter are not part of the official client API and have been removed from MySQL 5.7.9. Use mysql_get_option instead.

### DIFF
--- a/mysql.xs
+++ b/mysql.xs
@@ -872,6 +872,9 @@ dbd_mysql_get_info(dbh, sql_info_type)
     IV type = 0;
     SV* retsv=NULL;
     bool using_322=0;
+#if MYSQL_VERSION_ID >= 50709
+    IV net_buffer_length;
+#endif
 
     if (SvMAGICAL(sql_info_type))
         mg_get(sql_info_type);
@@ -903,6 +906,9 @@ dbd_mysql_get_info(dbh, sql_info_type)
 	    retsv = newSVpv(!using_322 ? "`" : " ", 1);
 	    break;
 	case SQL_MAXIMUM_STATEMENT_LENGTH:
+#if MYSQL_VERSION_ID >= 50709
+        mysql_get_option(NULL, MYSQL_OPT_NET_BUFFER_LENGTH, &net_buffer_length);
+#endif
 	    retsv = newSViv(net_buffer_length);
 	    break;
 	case SQL_MAXIMUM_TABLES_IN_SELECT:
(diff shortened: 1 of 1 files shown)


mbeijen:
Hi berntm, I guess MySQL 5.7.9 is not yet available right? The last version I can find on the MySQL website is 5.7.8rc.

What is the earliest version of MySQL where the mysql_get_option alternative could be used? Maybe we can change the code to use `mysql_get_option` not just on 5.7.9 but earlier versions as well.

berntm:
Hi. You're correct that 5.7.9 is not available yet.
mysql_get_option is not new, but MYSQL_OPT_NET_BUFFER_LENGTH which is used to retrieve the needed value is introduced in 5.7.9.

mbeijen:
So if I'm not mistaken you remove a feature (undocumented, but used, for a long time apparently), and in *the same version* you introduce functionality to replace this? No depreciation cycles or nothing?

And while I appreciate you making the pull request, it's strange to add a feature for open source software that is not even released, don't you think?

Not to mention that with MariaDB existing and using a different numbering scheme, checking for version number as you do in this patch can have impact on the *other* databases.

So basically I'd like to ask you to:

1. make a longer depreciation cycle for removal of this feature, preferably for the NEXT major MySQL release.
2. think about a patch that uses option checking (e.g. "is there `MYSQL_OPT_NET_BUFFER_LENGHT`, if so, use that, otherwise use the old way")

I'd be more than happy to accept a patch like that!

berntm:
MySQL 5.7.9 will be out soon. I wanted to have a patch for DBD-mysql ready. The patch should of course not be merged to DBD-perl before 5.7.9 source is publicly available.

Changed the check to avoid possible MySQL vs forks version number confusion.

mbeijen:
This looks better, thanks.

berntm:
There is already a report on this: http://bugs.mysql.com/bug.php?id=78926 so a merge into DBD-mysql would be appreciated.

candosys:
What is the solution to this problem, obviously someone has not followed the release procedure; whoever that was, you've eaten up all my Saturday and Sunday free time

mbeijen:
Hi @candosys, I don't fully feel responsible for your lost time; mysql felt like they needed to remove a feature and replace it with another within one release and without depreciation.
It's now fixed in git and will be on CPAN shortly. 

@berntm thanks for the patch!

Sorry for the inconvenience caused!

candosys:
Thanks Michiel,


I took the fix in the git...thanks after 48 hrs, I finally stumbled on your
fix.

Thanks.

PR #44, opened 2015-11-10
Document that bracketed IPv6 is required

grypyrg:
Document that bracketed IPv6 is required.


mbeijen:
Thanks for your contribution!

PR #45, opened 2015-11-14
Fix use after free error.

hannob:
When my_login fails the code tries to call mysql_errno on the mysql connection. However my_login has already free'd that connection variable, therefore causing a use-after-free error.

This patch changes that so that the free happens after the call to the error functions.

This was found with the help of Address Sanitizer.


mbeijen:
Thanks!

hannob:
thanks for merging, however can we have a new release to get this out soon? (I think use after free's can be serious and potentially lead to malfunction and security issues)

mbeijen:
I just put a test release on CPAN!

PR #46, opened 2015-11-20
Mention how to make the module have SSL enabled

genio:
I, admittedly stupidly, ran around in circles when trying to figure out why SSL connections weren't working properly on a Windows box with a brand new install of Strawberry Perl 5.22.

Going through that made me realize that I might have realized the error of my ways slightly faster if it was documented.

Thanks,
Chase

### DIFF
--- a/lib/DBD/mysql/INSTALL.pod
+++ b/lib/DBD/mysql/INSTALL.pod
@@ -519,6 +519,26 @@ you may also try to link without gzip libraries.
 
 =back
 
+=head1 ENABLE ENCRYPTED CONNECTIONS
+
+Connecting to your servers over an encrypted connection (SSL) is easily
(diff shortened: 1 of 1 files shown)


mbeijen:
Thanks for the documentation, it's helpful; I just merged it.
I think we might want to consider enabling SSL by default, what would you think?

genio:
The idea of having it on by default would get a +1 from me.

Does that cause problems of extra requirements at build time?  If so, I'd be -1 if the requirements chain gets too cumbersome.

mbeijen:
only libssl-dev. Also, in MySQL 5.7 SSL is now enabled by default on the server, if SSL header files are found during compilation. This would probably (hopefully?) lead to more SSL adoption for MySQL connections.

PR #47, opened 2015-11-20
RaiseError 1 causing make test to FAIL on machines without a mysql available for testing

klkane:
I have noticed that one machines that do not have a mysql db available for testing rather than make test skipping the tests and issuing a PASS so the module will install it results in a FAIL and refuses to install without intervention.  

RaiseError => 0 does not exhibit this behavior and results in no change in behavior when there is in fact a DB available for testing.  

This will restore the ability to install DBD::mysql without having a mysql db for the tests to pass.

### DIFF
--- a/t/05dbcreate.t
+++ b/t/05dbcreate.t
@@ -11,7 +11,7 @@ require 'lib.pl';
 
 my $dbh;
 eval {$dbh= DBI->connect('DBI:mysql:', $test_user, $test_password,
-                      { RaiseError => 1, PrintError => 1, AutoCommit => 0 });};
+                      { RaiseError => 0, PrintError => 1, AutoCommit => 0 });};
 if ($@) {
     diag $@;
     plan skip_all => "no database connection";


mbeijen:
Thanks for your PR. A while back I tried to make sure the tests pass if there is no database.
But for me your patch is not needed, and actually it does *not* pass if I run the test suite when I merge the patch.

I tried on a couple of different environments.

Can you please share details of your setup?

Below test results when I use your patch:

```
t/05dbcreate.t .. DBI connect('','root',...) failed: Can't connect to local MySQL server through socket '/var/run/mysqld/mysqld.sock' (2) at t/05dbcreate.t line 13.

1..3
not ok 1 - Connected to database
not ok 2 - CREATE DATABASE IF NOT EXISTS

#   Failed test 'Connected to database'
#   at t/05dbcreate.t line 21.

#   Failed test 'CREATE DATABASE IF NOT EXISTS'
#   at t/05dbcreate.t line 23.
# Can't call method "do" on an undefined value at t/05dbcreate.t line 22.
Can't call method "disconnect" on an undefined value at t/05dbcreate.t line 26.
# Looks like you planned 3 tests but ran 2.
# Looks like you failed 2 tests of 2 run.
# Looks like your test exited with 255 just after 2.
Dubious, test returned 255 (wstat 65280, 0xff00)
Failed 3/3 subtests 

Test Summary Report
-------------------
t/05dbcreate.t (Wstat: 65280 Tests: 2 Failed: 2)
  Failed tests:  1-2
  Non-zero exit status: 255
  Parse errors: Bad plan.  You planned 3 tests but ran 2.
Files=1, Tests=2,  0 wallclock secs ( 0.01 usr  0.00 sys +  0.04 cusr  0.00 csys =  0.05 CPU)
Result: FAIL
```

klkane:
Ahh thats my bad, I screwed up totally what I thought the problem was.  If you do a vanilla perlbrew with 5.20.0 and install DBI.  DBD::mysql will fail and it seems its because Test::Deep is required from what I can gather.

I install Test::Deep and the problem goes away, I will keep looking at this and maybe give you a pull request that actually solves the correct problem I am seeing.

mbeijen:
Well, Test::Deep is specified as a build-time-dependency via the `Makefile.PL`  and also properly indicated in the dists metadata: https://metacpan.org/source/CAPTTOFU/DBD-mysql-4.033/META.json - so the behaviour you describe there is as intended.

PR #49, opened 2016-06-22
Added missing preprocessor

ikari7789:
The "if" is started inside of the preprocessor on line 3612, but closed outside of the preprocessor causing an error.

### DIFF
--- a/dbdimp.c
+++ b/dbdimp.c
@@ -1713,6 +1713,7 @@ MYSQL *mysql_dr_connect(
           mysql_options(sock, MYSQL_OPT_CONNECT_TIMEOUT,
                         (const char *)&to);
         }
+#if (MYSQL_VERSION_ID >= 40101)
         if ((svp = hv_fetch(hv, "mysql_write_timeout", 19, FALSE))
             &&  *svp  &&  SvTRUE(*svp))
         {
@@ -1745,6 +1746,7 @@ MYSQL *mysql_dr_connect(
                           " secure auth\n");
           mysql_options(sock, MYSQL_SECURE_AUTH, &secauth);
         }
+#endif
         if ((svp = hv_fetch(hv, "mysql_read_default_file", 23, FALSE)) &&
             *svp  &&  SvTRUE(*svp))
         {
@@ -3640,7 +3642,9 @@ int dbd_st_execute(SV* sth, imp_sth_t* imp_sth)
         return 0;
     }
 #endif
+#if MYSQL_VERSION_ID >= SERVER_PREPARE_VERSION
   }
+#endif
 
   if (imp_sth->row_num+1 != (my_ulonglong)-1)
   {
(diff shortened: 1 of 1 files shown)


mbeijen:
Hi Ikari,
I think you're right - but it looks like the code has been broken since a commit in 2005.
https://github.com/perl5-dbi/DBD-mysql/commit/7e37731917b3816235855b0576efd70231b63f56

also I found:

    #define SERVER_PREPARE_VERSION 40103

So are you really trying to compile against libraries OLDER than 4.1.3?
(Remember, 4.1 was the release that added subquery support!)

Actually, @CaptTofu and I had been discussing ripping out support for versions < 5 which cleans up the code base a lot. I think that might be the best solution here. And in the case you actually need to compile with 3.x libs or so you can also still compile against a 2.x version of DBD::mysql.

Please let me know what you think!

ikari7789:
Yes, unfortunately I'm currently stuck with MySQL 4.0.27 in our production environment. I know we should update, but it's easier to update this source code than updating our DBs.

I'm making another fix to this. I found another offender in testing (I tested this patch with an older version of DBD::mysql at first). I also have another PR that I'll be making for another location as well.

dbdimp.c
```diff
@@ -3657,8 +3661,10 @@ int dbd_st_execute(SV* sth, imp_sth_t* imp_sth)
       /** Store the result in the current statement handle */
       DBIc_NUM_FIELDS(imp_sth)= mysql_num_fields(imp_sth->result);
       DBIc_ACTIVE_on(imp_sth);
+#if MYSQL_VERSION_ID >= SERVER_PREPARE_VERSION
       if (!imp_sth->use_server_side_prepare)
         imp_sth->done_desc= 0;
```
pali: I think you want to set done_desc to 0 even if server side prepare statements are not supported by mysql client...

pali:
I think that this patch is small to have problems with maintaining... And if there are still people interested in usage of new DBD::mysql with old mysql client library I do not see problem with taking such really small patch to help them...

CaptTofu:
Hi all,

At Kubecon, just replying now:


> On Jun 21, 2016, at 11:22 PM, Michiel Beijen <notifications@github.com> wrote:
> 
> Hi Ikari,
> I think you're right - but it looks like the code has been broken since a commit in 2005.
> 7e37731 <https://github.com/perl5-dbi/DBD-mysql/commit/7e37731917b3816235855b0576efd70231b63f56>
> also I found:
> 
> #define SERVER_PREPARE_VERSION 40103
> So are you really trying to compile against libraries OLDER than 4.1.3?
> (Remember, 4.1 was the release that added subquery support!)
> 
> Actually, @CaptTofu <https://github.com/CaptTofu> and I had been discussing ripping out support for versions < 5 which cleans up the code base a lot. I think that might be the best solution here. And in the case you actually need to compile with 3.x libs or so you can also still compile against a 2.x version of DBD::mysql.
> 
> 

I agree both with the idea of cleaning up the driver and your (Michiel) recommendation. That’s a really ancient version of MySQL (I started working at MySQL in 2004 around that version!) so the path of least pain if you have to use it is use a stable release that worked with 4.0.x

CaptTofu:
Sorry, for some reason this old email somehow showed up in a much more recent thread that I was addressing.

My mail client frustrates me.

mbeijen:
On Wed, Nov 9, 2016 at 2:04 AM, Patrick Galbraith <notifications@github.com>
wrote:

> Sorry, for some reason this old email somehow showed up in a much more
> recent thread that I was addressing.
>
> My mail client frustrates me.


I thought you *LOVED* Emacs? :D

--
Michiel

mbeijen:
This was fixed in 3f0b37e4fc47d91c81c54c5a50b94c6307704e1d -  thanks!!

PR #50, opened 2016-07-01
Check for errors after calling mysql_affected_rows()

pali:
Function mysql_affected_rows() returns 64 bit unsigned integer and highest
value means an error occurred. Handle return value correctly as -2.

When two concurrent calls were trying to acquire same row lock then DBD
returned nonsense error message "Statement has no result columns to bind"
or "fetch() without execute()". And DBI did not called DBD error handler
code ref. See bug reports:

https://rt.cpan.org/Public/Bug/Display.html?id=75353
https://rt.cpan.org/Public/Bug/Display.html?id=83351
https://bugs.mysql.com/bug.php?id=64475

After fixing check for return value of mysql_affected_rows then DBD returns
true mysql error "Lock wait timeout exceeded; try restarting transaction"
and also calls error handler code ref correctly.

### DIFF
--- a/dbdimp.c
+++ b/dbdimp.c
@@ -3396,13 +3396,15 @@ my_ulonglong mysql_st_internal_execute(
             mysql_use_result(svsock) : mysql_store_result(svsock);
 
           if (mysql_errno(svsock))
-            do_error(h, mysql_errno(svsock), mysql_error(svsock)
-                     ,mysql_sqlstate(svsock));
-
-          if (!*result)
-            rows= mysql_affected_rows(svsock);
-          else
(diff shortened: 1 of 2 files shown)


mbeijen:
Thanks for the PR, I merged it here!
https://github.com/perl5-dbi/DBD-mysql/commit/30acbede23709235bc8b003b6f5a4c1eb93c628e

I also released https://metacpan.org/changes/release/MICHIELB/DBD-mysql-4.033_03 which has the patch and will release 4.034 (non-dev) when there is no strange feedback from CPANTesters.

PR #52, opened 2016-07-15
Add $dbh->ping() to docs

dveeden:
Not sure if this really belongs in the docs... so feel free to change it or drop it alltogether

### DIFF
--- a/lib/DBD/mysql.pm
+++ b/lib/DBD/mysql.pm
@@ -1434,6 +1434,12 @@ important if you modify access privileges or create new users.
 
 =back
 
+=item ping
+
+This can be used to send a ping to the server.
+
+    $rc = $dbh->ping();
+
 =back
 
 


mbeijen:
Hi Daniël, 

Your PR let me look into what DBD::mysql's ping() actually does.

I found it calls [mysql_ping()](https://dev.mysql.com/doc/refman/5.7/en/mysql-ping.html) which internally does a reconnect if needed. But, maybe for historical reasons, or maybe for other reasons, if `mysql_ping()` fails, in `mysql.xs`, we ALSO do a `mysql_db_reconnect()` -- see line https://github.com/perl5-dbi/DBD-mysql/blob/4.034/mysql.xs#L580

I'd want to remove the call to `mysql_db_reconnect()` in order to prevent DBD::mysql from performing unnecessary extra work. What would you think about that?

dveeden:
I tried this:
```
#!/usr/bin/perl

use strict;
use warnings;
use DBI;

my $dbh = DBI->connect("DBI:mysql:database=test;host=[::1];port=5712",
                       "msandbox", "msandbox",
                       {'RaiseError' => 1, 'mysql_auto_reconnect' => 1});

print("DO 1\n");
$dbh->do("DO 1");

sleep(5);

print("ping\n");
$dbh->ping();

sleep(5);

print("DO 1\n");
$dbh->do("DO 1");

sleep(5);

print("ROLLBACK RELEASE\n");
$dbh->do("ROLLBACK RELEASE");

sleep(5);

print("ping\n");
$dbh->ping();

sleep(5);

print("DO 1\n");
$dbh->do("DO 1");

sleep(5);

print("disconnect\n");
$dbh->disconnect();
```
screenshot from 2016-07-15 14-24-33

Seems like the $dbh->ping() doesn't actually do a reconnect, that's done on the next command ('DO 1' in this case).

The reconnect on ping failure seems indeed wasteful. Could there be any cases where MYSQL_OPT_RECONNECT is not set but we would like to reconnect anyways?

mbeijen:
Thanks for the doc patch. I merged it here:

https://github.com/perl5-dbi/DBD-mysql/commit/88692a52cf290224d7119b25df62a5e9947ca051

as for the reconnect on ping failure; would that be something you might have time to do some testing with? I'd wonder how your company would handle reconnection in general

--
Michiel

PR #53, opened 2016-09-18
Add support for BIT type and coerce scalar value to correct types when mysql_server_prepare=0

pali:
Without mysql_server_prepare=1 DBD::mysql returns stringified numbers (e.g. "1") for columns of INT type. Second patch fix this problem and coerce scalars to correct perl types.

Both patches adding support for MYSQL_TYPE_BIT. When value has max 32 bits then it is stored as perl unsigned integer. Otherwise as raw string. Address bug: https://rt.cpan.org/Public/Bug/Display.html?id=88006

### DIFF
--- a/dbdimp.c
+++ b/dbdimp.c
@@ -336,7 +336,6 @@ free_param(pTHX_ imp_sth_ph_t *params, int num_params)
   }
 }
 
-#if MYSQL_VERSION_ID >= SERVER_PREPARE_VERSION
 /* 
   Convert a MySQL type to a type that perl can handle
 
@@ -402,7 +401,6 @@ static enum enum_field_types mysql_to_perl_type(enum enum_field_types type)
   }
   return(enum_type);
 }
-#endif
 
 #if defined(DBD_MYSQL_EMBEDDED)
 /* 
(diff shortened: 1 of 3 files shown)


pali:
Looks like travis is broken as it skipped all tests with message:

```
t/05dbcreate.t ....................... # DBI connect('test','travis',...) failed: Unknown database 'test' at t/05dbcreate.t line 16.
t/05dbcreate.t ....................... skipped: no database connection
t/10connect.t ........................ # DBI connect('test','travis',...) failed: Unknown database 'test' at t/10connect.t line 14.
```

mbeijen:
Thanks for your pull request. It looks good!

I think I broke Travis and AppVeyor tests myself when I removed creating
test databases here:

https://github.com/perl5-dbi/DBD-mysql/commit/138d73f719774b5eefd2bfd4000aaaecbaf45822

I'll try look into that later today.

--
Michiel

Op zondag 18 september 2016 heeft pali <notifications@github.com> het
volgende geschreven:

pali:
Looks like t/05dbcreate.t is broken, because in that test is call `$dbh->do("CREATE DATABASE IF NOT EXISTS $test_db")`. Probably regex for 'remove database from DSN' in that file is incorrect.

mbeijen:
PR welcome :D

pali:
Easy, PR is here: https://github.com/perl5-dbi/DBD-mysql/pull/54

Btw, BIT conversion code is updated and simplified.

mbeijen:
Many thanks!

mbeijen:
Your fix seems to fail on MySQL 5.7: 
https://ci.appveyor.com/project/mbeijen/dbd-mysql-mf7me
Can you look into that?

pali:
That is probably windows-related... Is there any way to test some patch for that windows mysql 5.7 test suite?

mbeijen:
You can make a PR and it will run on AppVeyor.

And I ran the test suite on my Mac with 5.7 and that produces the same
results, so it is not Windows-related.

Op maandag 19 september 2016 heeft pali <notifications@github.com> het
volgende geschreven:

pali:
Now I see where could be problem...

Error message is:
`DBD::mysql::db do failed: Data too long for column 'flags' at row 3 at t/rt88006-bit-prepare.t line 34.`

Line 34 is:
`ok $dbh->do("INSERT INTO dbd_mysql_rt88006_bit_prep (id, flags) VALUES (1, b'10'), (2, b'1'), (3, b'1111111111111111111111111111111111111111')");`

And there is:
```
CREATE TEMPORARY TABLE `dbd_mysql_rt88006_bit_prep` (
`flags` bit(32) NOT NULL
```

Can you try to change `bit(32)` to `bit(64)` so it can store those 40 bits in above INSERT?

mbeijen:
Thanks! Of course that worked 👍

PR #55, opened 2016-09-19
Add support for 64bit types when perl is compiled with 64bit support

pali:
Perl's IV in scalar can store 64bit integer when perl was compiled with
64bit support (default on 64bit linux with gcc). Use this feature and
stores MYSQL_TYPE_LONGLONG and MYSQL_TYPE_BIT as integers instead strings
when possible.


pali:
This change should go to some devel release as it needs more testing on different platforms with different compilers...

mbeijen:
I typically release a devel release, I'll also write a message on the mailing list to sollicit for extra feedback.

PR #56, opened 2016-09-21
MySQL 8.0 adaptions

berntm:
1. Changes in SSL server verification
2. MySQL library that supports fabric is 6.x.x
3. mysql_shutdown has been deprecated since 5.7.9 and is now gone

### DIFF
--- a/dbdimp.c
+++ b/dbdimp.c
@@ -1901,7 +1901,7 @@ MYSQL *mysql_dr_connect(
 	    char *ca_path = NULL;
 	    char *cipher = NULL;
 	    STRLEN lna;
-#if MYSQL_VERSION_ID >= SSL_VERIFY_VERSION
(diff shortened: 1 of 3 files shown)


mbeijen:
Hi, for mysql_shutdown(), this is *deprecated* and not removed from 5.7.9. So that is why I removed the functions from the documentation but did not completely remove the code yet:

https://github.com/perl5-dbi/DBD-mysql/commit/e35a6284602900c4085d1f82d34756977684defb

With your code `shutdown` would still work, only using a query. That's kind of neat.

Why are the `reload`, `createdb` and `dropdb` functions that have been deprecated in 5.7 not been removed in 8?

berntm:
Hi,
Don't know, but I can check out the fate of `reload`/`creatdb`/`dropdb`.

PR #68, opened 2016-10-26
Fix compilation of embedded mysql server

pali:
Compilation of embedded mysql server is working. Embedded tests are passing too. Now travis-ci run tests for both normal and embedded server.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,8 +11,8 @@ perl:
   - "5.24"
 before_install:
  - sudo apt-get update -qq
- - sudo apt-get install -qq libmysqlclient-dev
+ - sudo apt-get install -qq libmysqlclient-dev libmysqld-dev libwrap0-dev
  - perlbrew install-cpanm -f
 install:
   - cpanm --quiet DBI Test::Pod Test::Deep Test::DistManifest Proc::ProcessTable
-script: "export RELEASE_TESTING=1 && perl Makefile.PL && make disttest"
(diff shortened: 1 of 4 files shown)


pali:
Hm... t/manifest.t is failing :-( Looks like it should ignore mysqlEmb directory.

mbeijen:
I fixed `t/manifest.t` by adding e932e1f7f6d3a179200195614acfd711a2d5fc9a to master.

mbeijen:
I've *never* worked with the embedded server. I think the complaint about the missing table is because you should first call `mysql_library_init`. 

http://dev.mysql.com/doc/refman/5.7/en/libmysqld-example.html

On the other hand, embedded use has been broken for a LONG time. Also, the manual states:

> **Note**
> The libmysqld embedded server library is deprecated as of MySQL 5.7.17 and will be removed in a future version of MySQL.

(ref: http://dev.mysql.com/doc/refman/5.7/en/libmysqld.html)
And, of course, sqlite!

So while I think it is great if we can get it fixed, another option might be to rip the code out altogether. Let me know what you think!

pali:
Current implementation is working. We just passed unrecognized server parameter --skip-dbd in tests. Which caused that problem "Embedded server was not started".

Warning message "mysql_embedded: Table 'mysql.plugin' doesn't exist" is OK here. We just do not have plugin support but that does not matter for simple tests.

PR #69, opened 2016-10-28
Commit 8e7ce0b "MYSQL_OPT_NET_BUFFER_LENGTH not defined in mariadb-10…

berntm:
Commit 8e7ce0b breaks compatability with MySQL. MYSQL_OPT_NET_BUFFER_LENGTH is not a macro but an enum, so we need to need to distinguish between MySQL and MariaDB.

### DIFF
--- a/mysql.xs
+++ b/mysql.xs
@@ -882,7 +882,8 @@ dbd_mysql_get_info(dbh, sql_info_type)
     IV type = 0;
     SV* retsv=NULL;
     bool using_322=0;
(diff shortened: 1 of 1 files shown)


berntm:
Note: This is verified for MySQL 5.7 and 8.0, but not verified for MariaDB

Tux:
Verified to work on mariadb-10.0.22-3.7
@mbeijen fine to merge!

PR #72, opened 2016-11-06
Add new attribute mysql_server_prepare_disable_fallback which disable any fallback from prepared statement to normal non-prepared

pali:
In some cases it is better to throw error that mysql server cannot prepare
or execute some statement as server-prepared instead fallbacking to normal
non-prepared one (and without any control from user application).

Now applications can set mysql_server_prepare_disable_fallback to 1 disable
this automatic fallback. Default value is 0 (= fallbacks are enabled).


pali:
circleci tests are failing :-(

PR #74, opened 2016-11-27
add fallback option to treat BIGINTs as strings

manzur:
For compatibility reasons it could useful to have the old
behaviour of interpreting BIGINTs as strings.

This could be useful for the following reasons:

* to support code that could potentially run on Perl with `64bitint`
option and without it;

* to be able to treat BIGINTs(from the same source) similarly across
different programming languages without a surprise for the cross-language developers;

* to support existing projects that encode their output(i.e in JSON) to be
processed by services that are strict on decoding format.

### DIFF
--- a/Makefile.PL
+++ b/Makefile.PL
@@ -53,7 +53,8 @@ Getopt::Long::GetOptions(
     "embedded=s",
     "mysql_config=s",
     "force-embedded",
-    "with-mysql=s"
+    "with-mysql=s",
+    "bigintasastring"
     ) || die Usage();
 
 my $source = {};
@@ -98,7 +99,7 @@ MSG
 
 for my $key (qw/testdb testhost testuser testpassword testsocket testport
                     cflags embedded libs nocatchstderr nossl nofoundrows
-                    ps-protocol bind-type-guessing force-embedded/)
+                    ps-protocol bind-type-guessing force-embedded bigintasastring/)
 {
   Configure($opt, $source, $key);
 }
@@ -240,6 +241,7 @@ if ($^O eq 'VMS') {
 $cflags .= " -DDBD_MYSQL_WITH_SSL" if !$opt->{'nossl'};
 $cflags .= " -DDBD_MYSQL_INSERT_ID_IS_GOOD" if $DBI::VERSION > 1.42;
 $cflags .= " -DDBD_NO_CLIENT_FOUND_ROWS" if $opt->{'nofoundrows'};
+$cflags .= " -DDBD_BIGINT_AS_A_STRING" if $opt->{'bigintasastring'};
 $cflags .= " -g ";
 my %o = ( 'NAME' => 'DBD::mysql',
 	  'INC' => $cflags,
@@ -627,7 +629,7 @@ perl Makefile.PL --testuser=username
       $opt->{$param} = $user;
 	  $source->{$param} = 'guessed';
     }
-    elsif ($param eq "nocatchstderr" || $param eq "nofoundrows") {
+    elsif ($param eq "nocatchstderr" || $param eq "nofoundrows" || $param eq 'bigintasastring') {
       $source->{$param} = "default";
       $opt->{$param} = 0;
     }
--- a/dbdimp.c
+++ b/dbdimp.c
@@ -359,7 +359,6 @@ static enum enum_field_types mysql_to_perl_type(enum enum_field_types type)
   case MYSQL_TYPE_INT24:
   case MYSQL_TYPE_YEAR:
 #if IVSIZE >= 8
-  case MYSQL_TYPE_LONGLONG:
     enum_type= MYSQL_TYPE_LONGLONG;
 #else
     enum_type= MYSQL_TYPE_LONG;
@@ -379,8 +378,10 @@ static enum enum_field_types mysql_to_perl_type(enum enum_field_types type)
     enum_type= MYSQL_TYPE_DECIMAL;
     break;
 
-#if IVSIZE < 8
   case MYSQL_TYPE_LONGLONG:
+#if IVSIZE >= 8 && !DBD_BIGINT_AS_A_STRING
+    enum_type= MYSQL_TYPE_LONGLONG;
+    break;
 #endif
   case MYSQL_TYPE_DATE:
   case MYSQL_TYPE_TIME:
@@ -4969,12 +4970,17 @@ int dbd_bind_ph(SV *sth, imp_sth_t *imp_sth, SV *param, SV *value,
       case SQL_SMALLINT:
       case SQL_TINYINT:
 #if IVSIZE >= 8
-      case SQL_BIGINT:
           buffer_type= MYSQL_TYPE_LONGLONG;
 #else
           buffer_type= MYSQL_TYPE_LONG;
 #endif
           break;
+
+#if IVSIZE >= 8 && !DBD_BIGINT_AS_A_STRING
+      case SQL_BIGINT:
+          buffer_type= MYSQL_TYPE_LONGLONG;
+	  break;
+#endif
       case SQL_DOUBLE:
       case SQL_DECIMAL: 
       case SQL_FLOAT: 


mbeijen:
Hi, thanks for your patch. We also have this guy who reported he was using the MySQL option `smallint(5) unsigned zerofill` which is an option that blows my mind, it produces `INT`s that look like `00001`.  (see https://rt.cpan.org/Ticket/Display.html?id=118977)

He might also have benefit of using a fallback mechanism.

But IF we'd do some sort of fallback mechanism it should not be a compile time option I think, but rather an option you can pass in the connection string. That way you can only enable it in scripts that really need this kind of fallback. What do you think?

And @pali @CaptTofu  : what are your thoughts?

manzur:
> But IF we'd do some sort of fallback mechanism it should not be a compile time option I think, but rather an option you can pass in the connection string. That way you can only enable it in scripts that really need this kind of fallback. What do you think?

Yes, it's possible to implement it as an attribute. Especially, it's more suitable for the first case: supporting same behaviour across different Perl versions. The only concern I had was that it needs more tweaking, and can impact performance. But if there's a consensus on moving this to runtime, I can work it as well.

pali:
Basically numeric value in perl scalar can be stored in 3 different ways: as integer; as double; as string. And perl does not distinguish between these three store types... number "3" is same as 3. So trying to distinguish between them in perl is wrong (e.g. by svflags)! And moreover perl doing automatic lazy coercion, so such code is immediately broken when you use perl.

Look at this code, calling length on $v automatically store string representation of integer 12 to $v -- even if branch is not called!
```
use Devel::Peek;
my $v = 12;
Dump $v;
if (!length($v)) { warn "do_something"; }
Dump $v;
```

But if you really want to get 64bit integers with POK without IOK, I will try to think how to do it correctly... This patch is not good and even not correct.

PS: https://rt.cpan.org/Ticket/Display.html?id=118977 looks like different problem, here mysql returns number formatted as string (format = with leading zeros) and this really must be stored as string, it is not number anymore.

manzur:
> Look at this code, calling length on $v automatically store string representation of integer 12 to $v -- even if branch is not called!

Yep, I'm aware of this.

> So trying to distinguish between them in perl is wrong (e.g. by svflags)! And moreover perl doing automatic lazy coercion, so such code is immediately broken when you use use perl.

No one is arguing with that. However, there're cases(posted above) like encoding your data to be used outside Perl. In that case if you don't specify each of your type explicitly, BIGINTs that used to be encoded as strings, will be encoded as integers, which can potentially break process decoding that output.

pali:
As perl does not distinguish between int <--> string you need explicitly specify type if you are going to convert perl scalar into external data (like JSON)... If you are not doing it, then you have a problem. (I know, JSON::PP and JSON::XS are good candidates which violate it... and should not be used for serialization)

But as I wrote, I will think how to implement such thing... It looks like ideal solution would be attribute (on statement and also in dbh) which return (internally) stringified number instead normal number for 64bit mysql int type.

pali:
Mhm... Now I do not it is a good idea to have such support (compile time or runtime). When perl is compiled with 64bit long integers, then perl's integers should be used for storing integers -- not strings. I would say it is bug when **native** integers are stored internally as strings.

Comments for your reasons:
>  to support code that could potentially run on Perl with 64bitint option and without it;

I guess, perl code which depends on size of perl integer is not very portable...

> to be able to treat BIGINTs(from the same source) similarly across different programming languages without a surprise for the cross-language developers;

BIGINT is 64bit long integer and I bet for developers is it common to use (64bit) integer type for their storage, not string storage (even type of storage is internal)

> to support existing projects that encode their output(i.e in JSON) to be processed by services that are strict on decoding format.

perl do not have strict types (like other languages) and such thing cannot be represented without objects in perl or without some external information.

This is just my opinion. I will let final decision to @mbeijen or @CaptTofu. What do you think now (after my comments)?

If your code depends on internal things like svflags, then such code is even not portable between different perl versions. And if you really need to have internal string of some scalars, I would rather suggest you to explicitly enforce that, instead trying to implement and use DBI attribute.

mbeijen:
I'm in agreement with @pali here - I'm sorry for the inconvenience - you were relying on a bug that BIGINTs were treated as strings. We'll not be bringing this faulty behaviour back to DBD::mysql.

ribasushi:
> you were relying on a bug that BIGINTs were treated as strings

Scenario:
 - User has an application that deals with MySQL BigInts ( something as simple as *storing file sizes* is sufficient to require it )
 - User runs a 32bit IV perl (quite common on low-end devices / VPSes )
 - User needs to upgrade their `DBD::mysql` in order to patch up a recent vulnerability

What does this user do now...?

pali:
* ad security & vulnerabilities: stable tls distributions just backport security fixes to older versions, so nothing other is changes
* perl with 32bit IVs internally stores mysql bigints as strings

If your application depends on internal flags and internal storage of perl structures then your application is not portable and suitable for upgrading perl or other modules...

So as long term solution I would suggest to fix application to not depend on internal storage.

As short term solution for this one problem I would suggest to wrap any fetch output by simple stringification call, eg. `map { "$_" }`

ribasushi:
I might be completely wrong, but the current behavior also seems to affect `SELECT`s, i.e. depending on `ivsize` `DBD::mysql` now returns different things on a simple `selectall_arrayref`. My understanding was that this is what the OP had a problem with in the first place...

pali:
@ribasushi Now I do not understand you... Please provide examples which returns different things.

ribasushi:
`$dbh->selectall_arrayref('SELECT some_bigint_field FROM sometable')`

@pali ^^ what does this return on 32bit and on 64bit IVs after your changes?

pali:
Before changes both returned reference to array with one scalar. That scalar contained number and internally stored in PV slot (as string).

After changes on perl with 64bit IV that scalar contains number internally stored in IV slot. Perl with 32bit IV was not changes and number is still stored internally in PV slot.

ribasushi:
So, on 64bit IVs now people who pass the result to e.g. `JSON::XS` will get different output only due to an upgrade and nothing else. And there seems to be no solution for them other than *well... just whack "" on all the values*

I think this change/ticket closure is shortsighted, and @manzur's patch should be reconsidered (be it compile-time or even better flag-controllable at runtime)

pali:
And if I passed integer number to some typed serializer then I expect that it returns integer, not a string. And I think different behavior is a bug.

As I wrote if some application depends on internals of perl, then such application is not portable and must take care about it. I provided quick fix for those application via map. It is simpler and not introduce another "hacks" into driver just because some applications are not portable.

ap:
> I passed integer number to some typed serializer then I expect that it returns integer, not a string. And I think different behavior is a bug.

Yes, obviously. But what does that have to do with anything? You’re talking about what a serialiser should do with its input. The point in contention is that the input got changed, by a recent change to DBD::mysql, not what the serialiser should do with it. This change means that code which until now passed strings to a serialiser will now pass numbers instead, except if it happens to be running on a 32-bit perl (which you’ll need to keep in mind from now on).

This should definitely be controlled by a flag at runtime.

The argument “if you pass these values straight to a serialiser you need to mark their types anyhow” does not apply: there’s no way to output BIGINT values as numbers on a 32-bit *without* marking them as numbers explicitly anyway. So people who expected numeric values in the output already had to do something to force the serialiser to output them as numbers (if their serialiser looks at the SV flags for that). But people who needed strings didn’t. But due to this change, now they do.

pali:
> The point in contention is that the input got changed, by a recent change to DBD::mysql

I do not think so. Value from DBD::mysql was not changed. It still represent same number. See perlnumber at http://perldoc.perl.org/perlnumber.html

perlnumber:
> perl can internally represent numbers in 3 different ways: as native integers, as native floating point numbers, and as decimal strings.

perlnumber:
> forcing an argument into a particular form does not change the stored number

So in both cases DBD returned scalar in which was stored number. And changing form how is stored (either in PV or in IV block) does not change the stored number.

> This change means that code which until now passed strings to a serialiser will now pass numbers instead.

So I would say before DBD::mysql passed numbers too, not strings.

> But people who needed strings didn’t. But due to this change, now they do.

They only way how to achieve status "people do not do nothing" is to fully revert those DBD changes. Patch attached into this pull request adds compile time flag, which means that people *must* recompile DBD on its own. My proposal is less drastic (as opposite to recompilation) with wrapping their code into some simple `map { "$_" }` call.

ap:
> Value from DBD::mysql was not changed. It still represent same number.

Before that you said this:

> If I passed integer number to some typed serializer then I expect that it returns integer, not a string. And I think different behavior is a bug.

So which one is it?

And again, just now you said:

> In both cases DBD returned scalar in which was stored number. And changing form how is stored (either in PV or in IV block) does not change the stored number.

But before that you said this:

> I would say it is bug when **native** integers are stored internally as strings.

Either the internal form is irrelevant or it’s relevant. It can’t be both at once. Which one is your position?

> They only way how to achieve status "people do not do nothing" is to fully revert those DBD changes.

That’s not the only way. You already mentioned the other way:

> It looks like ideal solution would be attribute (on statement and also in dbh) which return (internally) stringified number instead normal number for 64bit mysql int type.

… as long as that defaults to disabled. Then users who upgrade don’t need to do anything.

pali:
> > Value from DBD::mysql was not changed. It still represent same number.

> Before that you said this:

> > If I passed integer number to some typed serializer then I expect that it returns integer, not a string. And I think different behavior is a bug.

> So which one is it?

First quoted text was mean for DBD::mysql, second for any typed serializer (e.g. JSON).

Just that perl do not have strict type and so serializer must receive additional information to know if scalar is a string or an number. JSON::XS is not doing it and I think it is bug (or known feature, as authors define it).

> But before that you said this:

>> I would say it is bug when *native* integers are stored internally as strings.

> Either the internal form is irrelevant or it’s relevant. It can’t be both at once. Which one is your position?

Here I wrote about *native* integers and how are stored *internally*. For pure perl applications it should be irrelevant, but for C DBD::mysql code it should not be for different reasons (effectivity, numeric complexity, space storage, ...).

> > They only way how to achieve status "people do not do nothing" is to fully revert those DBD changes.

> That’s not the only way. You already mentioned the other way:

> > It looks like ideal solution would be attribute (on statement and also in dbh) which return (internally) stringified number instead normal number for 64bit mysql int type.

> … as long as that defaults to disabled. Then users who upgrade don’t need to do anything.

In following comments (https://github.com/perl5-dbi/DBD-mysql/pull/74#issuecomment-263555508) I realized it is not a good idea.

And having such option disabled by default I consider as a bug.

mjegh:
I don't want to get involved in the argument but I will point you are https://metacpan.org/pod/DBI#sql_type_cast which was a solution adopted years ago.

mjegh:
I forgot to say it is used in DBD::ODBC and DBD::Oracle.

pali:
@mjegh Hm.. that looks like a function which @manzur and @ap want. Can you show how DBD::ODBC and DBD::Oracle is using it?

mjegh:
https://metacpan.org/pod/DBD::ODBC#Additional-bind_col-attributes
https://metacpan.org/source/MJEVANS/DBD-ODBC-1.56/dbdimp.c#L4294
https://metacpan.org/source/MJEVANS/DBD-ODBC-1.56/dbdimp.c#L3464

It is similar in DBD::Oracle (which I did first) - just search for sql_type_cast and discardstring/strictly typed.

The background was were retrieving resultsets from Oracle which needed to be stored in JSON encoded files. Javascript knows something is an integer in JSON if it is unquoted and in any case we wanted the file size to be as small as possible so 1234 was preferable to "1234". The JSON encoder was JSON::XS (although we use the Cpanel version these days) and it knew to write 1234 instead of "1234" if the scalar didn't have a pv. Without it we were forced to do scalar + 0 on every integer in the result-set which was extremely annoying and time consuming.

frioux:
FWIW at ZipRecruiter this change will likely keep us from upgrading for a long, long time.

ribasushi:
@pali @mbeijen your inaction on this issue is now causing downstream to carry patches like this:
https://cvsweb.openbsd.org/cgi-bin/cvsweb/ports/databases/p5-DBD-mysql/patches/patch-dbdimp_c?rev=1.4&content-type=text/x-cvsweb-markup
( via https://github.com/perl5-dbi/DBD-mysql/issues/78#issuecomment-501582280 )

pali:
@ribasushi what you described is totally different problem and not related to this issue. Details about root (magical scalars) are in my comment https://github.com/perl5-dbi/DBD-mysql/issues/78#issuecomment-282425847 So move discussion there.

mbeijen:
I'm no longer working on DBD::mysql. But I think it is time for Amavisd to realize that BIGINTS are not necessarily always returned as strings (esp. if you use 'new' DBD::mysql or DBD::mariadb) So I see openbsd 'fixed' this by patching DBD::mysql but I think this is the wrong way around.

PR #76, opened 2016-12-10
Fix support for magic scalars

pali:
In every XS function process get magic only once.

Perl versions prior to 5.13.2 will be still broken because of missing _nomg
variants of perl functions.

### DIFF
--- a/MANIFEST
+++ b/MANIFEST
@@ -73,6 +73,7 @@ t/90utf8_params.t
 t/91errcheck.t
 t/99_bug_server_prepare_blob_null.t
 t/lib.pl
+t/magic.t
 t/manifest.t
 t/mysql.dbtest
 t/pod.t
--- a/dbdimp.c
+++ b/dbdimp.c
@@ -73,7 +73,7 @@ void get_param(pTHX_ SV *param, int field, bool enable_utf8, bool is_binary, cha
   STRLEN len;
   int is_utf8;
 
-  buf = SvPV(param, len);
+  buf = SvPV_nomg(param, len);
   is_utf8 = SvUTF8(param);
   if (enable_utf8 && !is_binary && !is_utf8 && str_is_nonascii(buf, len))
   {
@@ -107,7 +107,7 @@ void get_statement(pTHX_ SV *statement, bool enable_utf8, char **out_buf, STRLEN
   STRLEN len;
   int is_utf8;
 
-  buf = SvPV(statement, len);
+  buf = SvPV_nomg(statement, len);
   is_utf8 = SvUTF8(statement);
   if (enable_utf8 && !is_utf8 && str_is_nonascii(buf, len))
   {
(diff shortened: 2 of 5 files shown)


mbeijen:
To be clear: If you say "Perl versions prior to 5.13.2 will be still broken" you mean "if you have < 5.13.2 this code optimization is not used" am I right?
The tests run fine on all perl versions.

pali:
This is not an optimization, but a code fix. As get magic on scalars is called more times in XS functions which should not. This patch is fixing it, but only for >= 5.13.2. There are no tests in this pull request :-( I can write some, but needs more time. Fell free to let open this merge request if you are waiting for tests.

Versions < 5.13.2 stay unfixed, due to missing _nomg functions...

mbeijen:
I noticed there are no tests, and I don't think it's needed per se. My point was, the existing tests are still running (no regressions).
So I understand now that this is not an optimization but a fix - on the other hand the existing behavior DOES work, although technically it's incorrect. Is my understanding correct?

pali:
Yes!

I will try to prepare some tests with tied scalars, to check it.

pali:
I created simple test for tied magic.

pali:
Hm... it is not enough! Do not merge!

pali:
Ok, tests are now failing for perl 5.14 and older as expected. Due to fundamental problems in those perl versions, DBD::mysql cannot be fixed.

mbeijen:
Hi Pali,

I found this generates a LOT of test output on STDERR because the stupid Test::More does not do binmode utf8 the terminal: https://travis-ci.org/perl5-dbi/DBD-mysql/jobs/185160984#L385

If possible I would like to avoid cluttering the test output.

pali:
Looks like Test::More is buggy itself and cannot process wide characters in test description :-( What we can do is to process every test description and remove wide characters before passing it to Test::More.

But first question is: What to do with older perl versions? They really cannot be fixed.

mjegh:
my $tb = Test::More->builder;
binmode $tb->failure_output, ":utf8";
binmode $tb->todo_output,    ":utf8";

pali:
@mjegh: thanks! test updated.

pali:
@mbeijen: What to do with older perl version? I can see only two options:
1) Drop support for older perl version
2) Add SKIP blocks into those tests and probably describe brokeness in POD documentation

mbeijen:
At this point I really feel we should support back to 5.10 and 5.8 as DBI and DBD::Pg also do that at this moment. People using older OSes (think RHEL or CentOS 6, or UNIXes) might want to recompile DBD::mysql to use newer features. It would be nice if we don't force them to also switch perls. So option 1 is not an option for me. But I'm really OK with a SKIP block that tests `$^V`.

pali:
Ok, I added skip blocks. But appveyor is broken now :-(

pali:
@mbeijen PING

mbeijen:
Thanks! I merged it now. Appveyor is broken because of the choco package manager, not because of the tests.

pali:
When will be Appveyor fixed and working again?

mbeijen:
Well, it's not appveyor per se, it's the strawberry package inside the chocolatey package manager they use.
I'll see if I can get that fixed.

pali:
Ok, as github show for each commit big red cross it would be good to have it working again...

PR #77, opened 2016-12-18
Basic GTID tracking.

dveeden:
https://rt.cpan.org/Public/Bug/Display.html?id=119122
https://dev.mysql.com/doc/refman/5.7/en/mysql-session-track-get-first.html

This code is only tested at a *very* basic level.

- The test runs, but other tests fail on a GTID enabled server. For example: 
```
t/53comment.t ........................... DBD::mysql::db do failed: Statement violates GTID consistency: CREATE TEMPORARY TABLE and DROP TEMPORARY TABLE can only be executed outside transactional context.  These statements are also not allowed in a function or trigger because functions and triggers are also considered to be multi-statement transactions. at t/53comment.t line 32.
```
- This is specific for GTID tracking and doesn't to schema/state/variable tracking. This could be made in a more generic feature, but currently I think this more specific feature makes sense.
- This is tested against 5.7 and it *should* be possible to compile it against 5.6 and below (with loss of function)
- I did not test against MariaDB server
- I did not test compilation against MariaDB libraries
- It probably works against MariaDB: https://jira.mariadb.org/browse/MDEV-8931
- MariaDB has a different GTID implementation, so it might behave differently (e.g. test might fail because the MariaDB implementation has a shorter GTID)
- The test only runs against a server were this is enabled in the config, so this gets very limited testing. Unfortunately this can't be enabled dynamically, so we can't enable it during the test.

Example:
```perl
#!/usr/bin/perl
use v5.24.0;
use DBI;

my $dbh = DBI->connect("DBI:mysql:database=test;mysql_socket=/tmp/mysql_sandbox5717.sock",
                       "msandbox", "msandbox",
                       {'RaiseError' => 1});

$dbh->do('FLUSH PRIVILEGES');
say "GTID= $dbh->{'mysql_gtids'}";

$dbh->disconnect();
```

### DIFF
--- a/dbdimp.c
+++ b/dbdimp.c
@@ -2854,6 +2854,28 @@ SV* dbd_db_FETCH_attrib(SV *dbh, imp_dbh_t *imp_dbh, SV *keysv)
       result= sv_2mortal((newRV_noinc((SV*)hv)));
     }
 
+#if MYSQL_VERSION_ID >= 50708
+#if MYSQL_VERSION_ID != 60000 /* Connector/C 6.0.x */
+#ifndef MARIADB_BASE_VERSION
+/* SESSION_TRACK_GTIDS was added in commit c4f32d662 in MySQL 5.7.8-rc */
+  case 'g':
+    if (strEQ(key, "gtids"))
+    {
+      const char *data;
+      size_t length;
+      if (mysql_session_track_get_first(imp_dbh->pmysql, SESSION_TRACK_GTIDS, &data, &length) == 0)
+      {
+        result= sv_2mortal(newSVpvn(data, length));
+      }
+      else
+      {
+        result= &PL_sv_undef;
+      }
+    }
+    break;
+#endif
+#endif
+#endif
   case 'h':
     if (strEQ(key, "hostinfo"))
     {
--- a/lib/DBD/mysql.pm
+++ b/lib/DBD/mysql.pm
@@ -1728,6 +1728,11 @@ this attribute determines whether a I<fetchrow> will chop preceding
 and trailing blanks off the column values. Chopping blanks does not
 have impact on the I<max_length> attribute.
 
+=item mysql_gtids
+
+Returns GTID(s) if GTID session tracking is ensabled in the server via
+session_track_gtids.
+
 =item mysql_insertid
 
 If the statement you executed performs an INSERT, and there is an AUTO_INCREMENT
--- a/t/57trackgtid.t
+++ b/t/57trackgtid.t
@@ -1,0 +1,32 @@
+use strict;
+use warnings;
+
+use DBI;
+use Test::More;
+
+use vars qw($test_dsn $test_user $test_password);
+use lib 't', '.';
+require "lib.pl";
+
+my $dbh;
+eval{$dbh = DBI->connect($test_dsn, $test_user, $test_password,
+			    {RaiseError => 1});};
+
+if ($@) {
+    plan skip_all =>
+        "no database connection";
+}
+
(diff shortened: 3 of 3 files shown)


mbeijen:
Hi Daniël, thanks for the pull request.
I found this also does not compile against mysql-connector-c version 6.1.6 which are 5.6 libraries;

```
dbdimp.c:2859:58: error: use of undeclared identifier 'SESSION_TRACK_GTIDS'; did you mean 'SESSION_TRACK_SCHEMA'?
      if (mysql_session_track_get_first(imp_dbh->pmysql, SESSION_TRACK_GTIDS, &data, &length) == 0)
                                                         ^~~~~~~~~~~~~~~~~~~
                                                         SESSION_TRACK_SCHEMA
```

Maybe it is better to depend on if `SESSION_TRACK_GTIDS` is defined than to define on the version number.

dveeden:
I used the version number from https://dev.mysql.com/doc/refman/5.7/en/mysql-session-track-get-first.html which is for the function, but not for SESSION_TRACK_GTIDS. I've now fixed the version number. 

Note that Connector/C 6.1.6 is not 5.6, but 5.7..
```
$ rpm -qf /usr/include/mysql/mysql_com.h
mysql-connector-c-devel-6.1.6-1.el7.i686

$ grep -w MYSQL_VERSION_ID /usr/include/mysql/mysql_version.h 
#define MYSQL_VERSION_ID            50706
```

Changing this to check if SESSION_TRACK_GTIDS is defined might indeed be a better option.

pali:
Hi! Please rebase your patches on top of master branch. Travis CI is now configured to test DBD::mysql with different MySQL & MariaDB versions.

pali:
So... it does not compile with MariaDB client.

dbdimp.c
```diff
@@ -2854,6 +2854,24 @@ SV* dbd_db_FETCH_attrib(SV *dbh, imp_dbh_t *imp_dbh, SV *keysv)
       result= sv_2mortal((newRV_noinc((SV*)hv)));
     }
 
+#if (MYSQL_VERSION_ID >= 50708) && (MYSQL_VERSION_ID <= MARIADB_VERSION_10)
```
pali: If this feature is not supported by MariaDB client, then rather check that MARIADB_BASE_VERSION is not defined.

In future Oracle can decide that MySQL version could be increased to 10 or 11 and this code stop working.

CaptTofu:
can we move this along? Any objections?

dveeden:
What needs to be done to get this merged?

dveeden:
This might be a slightly better example:
```perl
#!/usr/bin/perl
use v5.24.0;
use DBI;

my $master_dbh = DBI->connect("DBI:mysql:database=test;host=mymaster", "myuser", "mypass");
my $slave_dbh = DBI->connect("DBI:mysql:database=test;host=myslave1", "myuser", "mypass");

# Write to the master
$master_dbh->do('SET SESSION session_track_gtids=OWN_GTID');
$master_dbh->do("INSERT INTO t1() VALUES()");
my $gtid = $master_dbh->{'mysql_gtids'};
$master_dbh->disconnect();
say "GTID= $gtid";

# Wait until the GTID hits the slave (or timeout)
my $sth = $slave_dbh->prepare("SELECT WAIT_FOR_EXECUTED_GTID_SET(?, 10)");
$sth->execute($gtid);
if ($sth->fetchrow_arrayref()->[0] != 0) {
  say "Got timeout";
}
$sth->finish();
$slave_dbh->disconnect();
```

CaptTofu:
I merged this this morning - time to get a release going.

PR #79, opened 2017-01-14
Call missing finish for prepared statement in test t/40numrows.t

pali:
Without it that test shows warning:
DBD::mysql::st DESTROY failed: MySQL server has gone away.


pali:
warning is still there :-( do not merge for now.

PR #80, opened 2017-01-14
Fix parsing configure libs from mysql_config --libs output in Makefile.PL

pali:
Libraries in mysql_config --libs output can be specified by library name
with -l prefix or by absolute path to library name without any prefix.
Parameters must start with a hyphen, so treat all options without leading
hyphen in mysql_config --libs output as libraries with full path.

Partially fix bug: https://rt.cpan.org/Public/Bug/Display.html?id=100898


mbeijen:
👍

PR #82, opened 2017-01-15
Add mysql_ssl_mode option

dveeden:
This is available since 5.7.11 and replaces various other ssl options

See also https://dev.mysql.com/doc/refman/5.7/en/secure-connection-options.html#option_general_ssl-mode

### DIFF
--- a/dbdimp.c
+++ b/dbdimp.c
@@ -3144,6 +3144,12 @@ dbd_st_prepare_sv(
     prepare_retval= mysql_stmt_prepare(imp_sth->stmt,
                                        statement,
                                        statement_len);
+
+    if (prepare_retval && mysql_db_reconnect(sth))
+        prepare_retval= mysql_stmt_prepare(imp_sth->stmt,
+                                           statement,
+                                           statement_len);
+
     if (DBIc_TRACE_LEVEL(imp_xxh) >= 2)
         PerlIO_printf(DBIc_LOGPIO(imp_xxh),
                       "\t\tmysql_stmt_prepare returned %d\n",
@@ -3710,6 +3716,8 @@ my_ulonglong mysql_st_internal_execute41(
                   num_params);
 
   execute_retval= mysql_stmt_execute(stmt);
+  if (execute_retval && mysql_db_reconnect(sth))
+    execute_retval= mysql_stmt_execute(stmt);
   if (DBIc_TRACE_LEVEL(imp_xxh) >= 2)
     PerlIO_printf(DBIc_LOGPIO(imp_xxh),
                   "\t\tmysql_stmt_execute returned %d\n",
--- a/.travis.yml
+++ b/.travis.yml
@@ -75,7 +75,7 @@ before_install:
       esac ;
       SANDBOX_FILE="$HOME/$(basename "$SANDBOX_URL")" ;
     elif [ "$DB" = "MariaDB" ]; then
-      SANDBOX_URL=https://downloads.mariadb.org/interstitial/mariadb-$VERSION/bintar-linux-x86_64/mariadb-$VERSION-linux-x86_64.tar.gz/from/http%3A//ftp.hosteurope.de/mirror/archive.mariadb.org/ ;
+      SANDBOX_URL=https://downloads.mariadb.com/MariaDB/mariadb-$VERSION/bintar-linux-x86_64/mariadb-$VERSION-linux-x86_64.tar.gz ;
       SANDBOX_FILE="$HOME/mariadb-$VERSION-linux-x86_64.tar.gz" ;
     elif [ -n "$DB" ]; then
       echo "Unsupported DB '$DB'"; exit 1 ;
--- a/t/15reconnect.t
+++ b/t/15reconnect.t
@@ -10,13 +10,14 @@ use lib 't', '.';
 require 'lib.pl';
 
 my $dbh;
+my $sth;
 eval {$dbh = DBI->connect($test_dsn, $test_user, $test_password,
   { RaiseError => 1, AutoCommit => 1})};
 
 if ($@) {
   plan skip_all => "no database connection";
 }
-plan tests => 8 * 2;
+plan tests => 13 * 2;
 
 for my $mysql_server_prepare (0, 1) {
 $dbh= DBI->connect("$test_dsn;mysql_server_prepare=$mysql_server_prepare;mysql_server_prepare_disable_fallback=1", $test_user, $test_password,
@@ -37,6 +38,18 @@ ok(!$dbh->{Active}, "checking for inactive handle");
 ok($dbh->do("SELECT 1"), "implicitly reconnecting handle with 'do'");
 
 ok($dbh->{Active}, "checking for reactivated handle");
+
+ok($dbh->disconnect(), "disconnecting active handle");
+
+ok(!$dbh->{Active}, "checking for inactive handle");
+
+ok($sth = $dbh->prepare("SELECT 1"), "prepare statement");
+
+ok($sth->execute(), "implicitly reconnecting handle with executing prepared statement");
+
+ok($dbh->{Active}, "checking for reactivated handle");
+
+$sth->finish();
 
 $dbh->disconnect();
 }


dbdimp.c
```diff
@@ -1930,12 +1934,31 @@ MYSQL *mysql_dr_connect(
 		*svp)
 	      cipher = SvPV(*svp, lna);
 
+	    if ((svp = hv_fetch(hv, "mysql_ssl_mode", 14, FALSE)) &&
+		*svp)
+            {
+              ssl_mode_str = SvPV(*svp, lna);
```
pali: ssl_mode_str is declared only when #if MYSQL_VERSION_ID >= SSL_MODE_VERSION

So this code needs to be #if-ed too.

dbdimp.h
```diff
@@ -49,6 +49,7 @@
 #define NEW_DATATYPE_VERSION 50003
 #define SSL_VERIFY_VERSION 50023
 #define SSL_LAST_VERIFY_VERSION 50799
+#define SSL_MODE_VERSION 50711
```
pali: Is this version correct for both mysql and mariadb?

@mbeijen looks like we need some travis tests for both mysql and mariadb ideally for more versions...

dveeden:
I fixed the ifdef issue.

MariaDB doesn't seem to support MYSQL_OPT_SSL_MODE and also seems to have deprecated mysql_options() in favor of mysql_optionsv(). However MySQL doesn't have mysql_optionsv(). Also MySQL has mysql_options4() which seems to be supported by MariaDB.

mbeijen:
So, instead of checking version numbers, can't you use a construct like `#ifdef SSL_MODE_PREFERRED` or such? That seems much more robust. Of course we can check for `MARIADB` and then it's own set of version numbers but then if I start MichielDB and we'd have yet another vendor to `#ifdef` it would get very nasty.

dveeden:
@mbeijen Yes you are right

pali:
In this PR https://github.com/perl5-dbi/DBD-mysql/pull/85 I proposed changes for travis to test pull request changes against different mysql versions... It still does not work. But if you can wait until it is working and merged you can then rebase changes on of it. And you will see if your changes are really compatible with all tested mysql versions.

dveeden:
Yes I rather wait for that PR and have some more testing

dbdimp.c
```diff
@@ -1932,10 +1936,30 @@ MYSQL *mysql_dr_connect(
...
 	    mysql_ssl_set(sock, client_key, client_cert, ca_file,
 			  ca_path, cipher);
-#if MYSQL_VERSION_ID >= SSL_VERIFY_VERSION && MYSQL_VERSION_ID <= SSL_LAST_VERIFY_VERSION
+
+#ifdef SSL_MODE_PREFERRED
+	    if ((svp = hv_fetch(hv, "mysql_ssl_mode", 14, FALSE)) &&
+		*svp)
+	    {
+	          ssl_mode_str = SvPV(*svp, lna);
+	      if (strncmp(ssl_mode_str, "disabled", 8) == 0)
```
pali: Why strncmp? Do you really want to accept string "disabledXXX"?

dbdimp.c
```diff
@@ -1932,10 +1936,30 @@ MYSQL *mysql_dr_connect(
...
+	        ssl_mode = SSL_MODE_DISABLED;
+	      if (strncmp(ssl_mode_str, "preferred", 9) == 0)
```
pali: `else if` as only one ssl_mode can be specified.

dbdimp.c
```diff
@@ -1932,10 +1936,30 @@ MYSQL *mysql_dr_connect(
...
+	      if (strncmp(ssl_mode_str, "required", 8) == 0)
+	        ssl_mode = SSL_MODE_REQUIRED;
+	      if (strncmp(ssl_mode_str, "verify_ca", 9) == 0)
+	        ssl_mode = SSL_MODE_VERIFY_CA;
+	      if (strncmp(ssl_mode_str, "verify_identity", 15) == 0)
+	        ssl_mode = SSL_MODE_VERIFY_IDENTITY;
+	    }
+	    mysql_options(sock, MYSQL_OPT_SSL_MODE, &ssl_mode);
+	    if (ssl_mode != SSL_MODE_DISABLED)
+	      client_flag |= CLIENT_SSL;
```
pali: This should not happen as when you specify `mysql_ssl=1` SSL should be already enabled.

lib/DBD/mysql.pm
```diff
@@ -1186,6 +1186,18 @@ flags:
...
+This turns on the MYSQL_OPT_SSL_VERIFY_SERVER_CERT connection option.
+
+Note that this is deprecated as of MySQL 5.7.11 and is removed in MySQL 8.0
+
+=item mysql_ssl_mode
+
+Can be set to 'disabled', 'preferred', 'required', 'verify_ca' or 'verify_identity'.
+Default is 'preferred'.
+
+Available from MySQL 5.7.11
```
pali: This config logic is bad :-(

When you set `mysql_ssl=1` then SSL is `required`. And when you set `mysql_ssl_verify_server_cert=1` then mode should be `verify_identity`.

Rather do not export low-level libmysqlclient flags and configs to high-level DBD::mysql directly, because there is place for specifying nonsense combination like "mysql_ssl=1;mysql_ssl_mode=disabled"... Also I have feeling that with different MySQL and MariaDB versions I would be able to set probably same SSL/TLS options but with different low-level libmysqlclient functions and flags. And I think that DBD::mysql should not export directly those low-level flags but rather wrap them into stable API and did all C client logic in its own code... Think about application developer who want to use DBD::mysql. He really do not want to add into his code different if/else cases for different MySQL and MariaDB versions to correctly set SSL options.

Option `mysql_ssl_verify_server_cert` is already supported by DBD::mysql, but documentation is missing. This should be fixed first and probably support this option also for MySQL 8.0 (emulated by DBD::mysql).

What about introducing option `mysql_ssl_prefer` (with default value true) which will be equivalent for `SSL_MODE_PREFERRED` instead whole option `mysql_ssl_mode`? Becaue `SSL_MODE_DISABLED` is equivalent for `mysql_ssl=0`, `SSL_MODE_REQUIRED` for `mysql_ssl=1`, `SSL_MODE_VERIFY_IDENTITY` for `mysql_ssl_verify_server_cert=1` and `SSL_MODE_VERIFY_CA` is probably not needed...

dveeden: I started to work on this because I wanted to use `SSL_MODE_VERIFY_CA`, and that wasn't possible in `DBD::mysql`. At that time I knew that Oracle MySQL was moving from `--ssl` and `--ssl-verify-server-cert` to `--ssl-mode`.
* [WL#8785: Support multi-state SSL option](https://dev.mysql.com/worklog/task/?id=8785)
* [WL#9091: Remove --ssl, --ssl-verify-server-cert client-side options](https://dev.mysql.com/worklog/task/?id=9091)

I sort of expected MariaDB to follow the same path. But I now know that MariaDB server will most likely not follow that path and keep using the 'old' variables.

So I think you're right and `DBD::mysql` should implement a stable API.

DBD::mysql | MySQL 5.7+ | MariaDB
----------------|-----------------|-------------
**ssl=0**, ssl-verify-server-cert=0 (default) | `SSL_MODE_DISABLED` | ssl=0, ssl-verify-server-cert=0
ssl=?, **ssl_prefer=1** (new), ssl-verify-server-cert=0 | `SSL_MODE_PREFERRED` | Not set anything and let it use the defaults? 
**ssl=1**, ssl-verify-server-cert=0 | `SSL_MODE_REQUIRED` | ssl=1
ssl=1, ssl-verify-server-cert=0, **mysql_ssl_ca_(file_path)=xxx** | `SSL_MODE_VERIFY_CA` | set ca_file/ca_path for `mysql_ssl_set()`
ssl=1, **ssl-verify-server-cert=1** | `SSL_MODE_VERIFY_IDENTITY` | `MYSQL_OPT_SSL_VERIFY_SERVER_CERT`

`SSL_MODE_VERIFY_CA` is in many cases the highest level which still works. The certificates which are [generated by default](https://dev.mysql.com/doc/refman/5.7/en/creating-ssl-rsa-files-using-mysql.html) have a CN which is similar to "MySQL_Server_5.7.16_Auto_Generated_CA_Certificate". And even if one would use custom certificates with a hostname then it might not be possible to enable hostname validation because of the [missing support for SubjectAltName](https://bugs.mysql.com/bug.php?id=68052) (also for [MariaDB](https://jira.mariadb.org/browse/MDEV-10594) ). So using a virtual-IP, loadbalancer etc. can make it impossible to enable this.

Another question is what the default should be. In Oracle MySQL the default seems to be 'preferred' (see https://dev.mysql.com/doc/refman/5.7/en/secure-connection-options.html)
I do think the default should eventually be set to `SSL_MODE_VERIFY_IDENTITY`/`verify-server-cert=1,ssl=1`. Similar to what browsers and `curl` does by default. 

So I would suggest to do this:
1. Add `ssl_prefer`. Keep disabled by default for backwards compatibility. Maybe change this to `ssl_prefer=1` at a later stage?
2. Set `SSL_MODE_VERIFY_CA` if a ca file is specified (probably already what is happening now, but this could make it more explicit.

What should we expect to happen with MariaDB and `mysql_ssl_preferred`?

With `DBD::mysql` and MySQL 5.6:
[`--ssl`](https://dev.mysql.com/doc/refman/5.6/en/secure-connection-options.html#option_general_ssl) was similar to `SSL_MODE_PREFERRED`, this was changed in MySQL 5.7.6 to be `SSL_MODE_REQUIRED`. So compiling against 5.7 might suddenly cause a different behaviour. In 5.6 and older there was no real option to force the use of SSL/TLS on the client.

pali: Thank you for investigation!

> 1. Add ssl_prefer. Keep disabled by default for backwards compatibility. Maybe change this to ssl_prefer=1 at a later stage?

Now I'm thinking if it is good idea to have enabled SSL by default. Basically lot of people are running MySQL either on localhost or in secure network where is encryption just wasting of resources. And those who needs encryption still explicitly request for it.

So I'm OK with `ssl_prefer=0` as default option.

Maybe I would suggest to throw error when both option `ssl` and `ssl_prefer` are specified.

> 2. Set SSL_MODE_VERIFY_CA if a ca file is specified (probably already what is happening now, but this could make it more explicit.

Yes make sense. If somebody specify certificate CA the only thing what can be done is to verify if CA matches.

> What should we expect to happen with MariaDB and mysql_ssl_preferred?

MariaDB does not really support this opportunistic encryption mode? In this case throw error. And probably report bug to MariaDB bug tracker.

> In 5.6 and older there was no real option to force the use of SSL/TLS on the client.

Then this should be in DBD::mysql documentation.


Anyway, I'm thinking if new attribute `ssl_prefer` is needed and if is better or even worse to introduce tri-state logic for `ssl`: `ssl=0` (always disabled; default), `ssl=1` (always enabled), `ssl=2` (prefer if available).

pali: For 5.6 and older... is not there option to verify if established connection encrypted (before sending username/password)?

pali: Seems that it is not possible to verify if connection is encrypted before sending mysql password :-(

According to https://dev.mysql.com/doc/refman/5.6/en/mysql-ssl-set.html it is possible to verify if connection is encrypted or not via `mysql_get_ssl_cipher(sock)` -- but only after `mysql_real_connect(sock, host, user, passwd, ...)` was called. Which could send password over non encrypted channel.

It is better than nothing. But should be written DBD::mysql documentation with big warning!

dveeden: This even has a name: [BACKRONYM](http://backronym.fail/).

pali: Nice... I (independently) rediscovered this problem.

pali: Now I'm reading about that BACKRONYM and it describe just downgrade attack.

But what I pointed is that if somebody start doing that downgrade attack then you detect that attack **after** you send mysql password over (downgraded) non secure network!

As I wrote, calling `mysql_get_ssl_cipher()` (to detect if somebody is doing downgrade MITM attack) can be done **after** `mysql_real_connect()` which already send password to other side.

PR #85, opened 2017-01-17
Run travis tests on different MySQL and MariaDB versions

pali:
Changes in this pull request:
* Remember mysql_config value for Makefile.PL
* Check if specified Makefile.PL settings are working
* Run travis tests on different MySQL and MariaDB versions
* Run travis tests also on perl versions with -Duseshrplib

____

Travis matrix contains different MySQL and MariaDB version, namely those
which are available in stable long term supported Linux distributions.

Config file is prepared for easily extending database server matrix once
new versions are released.

Travis downloads official MySQL or MariaDB binary tarball and uses
MySQL::Sandbox for starting database server. DBD::mysql is then compiled
against client provided by that tarball and pre-installed perl 5.20 by
Travis.

Other perl versions are tested against system pre-installed database
version as before.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,8 +6,11 @@ perl:
   - "5.14"
   - "5.16"
   - "5.18"
+  - "5.18-shrplib"
   - "5.20"
+  - "5.20-shrplib"
   - "5.22"
+  - "5.22-shrplib"
   - "5.24"
 matrix:
   include:
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,10 +9,100 @@ perl:
   - "5.20"
   - "5.22"
   - "5.24"
+matrix:
+  include:
+    - perl: "5.20"
+      env: DB=MySQL VERSION=4.1.22
+    - perl: "5.20"
+      env: DB=MySQL VERSION=5.0.15
+    - perl: "5.20"
+      env: DB=MySQL VERSION=5.0.96
+    - perl: "5.20"
+      env: DB=MySQL VERSION=5.1.30
+    - perl: "5.20"
+      env: DB=MySQL VERSION=5.1.72
+    - perl: "5.20"
+      env: DB=MySQL VERSION=5.5.8
+    - perl: "5.20"
+      env: DB=MySQL VERSION=5.5.47
+    - perl: "5.20"
+      env: DB=MySQL VERSION=5.5.54
+    - perl: "5.20"
+      env: DB=MySQL VERSION=5.6.10
+    - perl: "5.20"
+      env: DB=MySQL VERSION=5.6.30
+    - perl: "5.20"
+      env: DB=MySQL VERSION=5.6.35
+    - perl: "5.20"
+      env: DB=MySQL VERSION=5.7.8-rc
+    - perl: "5.20"
+      env: DB=MySQL VERSION=5.7.17
+    - perl: "5.20"
+      env: DB=MySQL VERSION=8.0.0-dmr
+    - perl: "5.20"
+      env: DB=MariaDB VERSION=5.5.40
+    - perl: "5.20"
+      env: DB=MariaDB VERSION=5.5.47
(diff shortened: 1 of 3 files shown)


pali:
(testing, so do not merge)

mbeijen:
I wanted to say I'm very happy by you trying to do this! It has been on my wishlist for a while.
I guess I told you before, I'm really not concerned about MySQL version 4 at all. If you can fix the test failure, great, but I would not mind you dropping it off your matrix either. It's SO old!

pali:
Apparently that 4.1 test in matrix show problems in driver code... so it was useful!

pali:
There are 3 problems:
* Either Devel::CheckLib does not work on Windows or AppVeyor configuration is incorrect
* DBD::mysql cannot be compiled with MariaDB 10.2.3 because MariaDB provides incorrect mysql_config binary which says to link with -lmysqlclient, but it does not provide any libmysqlclient.so (just libmariadbclient.so)
* Test t/rt50304-column_info_parentheses.t does not pass with MySQL 8.0.0 and I have no idea why

pali:
First problem is not in Devel::CheckLib nor AppVeyor, but in windows version of mysql.h. It used type SOCKET without including winsock.h, so including pure mysql.h cause compiler error: `mysql.h:63:19: error: unknown type name 'SOCKET'`. Going to fix it manually in Makefile.PL.

pali:
Third problem with test t/rt50304-column_info_parentheses.t and MySQL 8.0.0 is fixed. DESCRIBE did not function with temporary tables (like in older MySQL 5 versions).

pali:
Looks like last MariaDB version with working mysql_config script is 10.2.1. As mysql_config is required for automatic compilation all new versions with broken mysql_config cannot be used for DBD::mysql... I will remove new versions from matrix.

mbeijen:
Hi Pali: *why* is the `mysql_config` script broken for MariaDB?

There seem to be lots of bugs for `mysql_config` in the MariaDB tracker:

https://jira.mariadb.org/browse/MDEV-9388?jql=text%20~%20%22mysql_config%22

pali:
Why? Because `mysql_config --libs` (from official tarball) outputs `-lmysqlclient` but in official mariadb binary tarballs (versions 10.2.2 and new) there is no `libmysqlclient.so` library. There is only `libmariadb.so` and `libmariadbclient.a` libraries.

pali:
You can look and compare build logs for MariaDB 10.2.1 (which is working fine) and MariaDB 10.2.2/10.2.3 (which fails in Makefile.PL):

10.2.1: https://travis-ci.org/perl5-dbi/DBD-mysql/jobs/193932892 (OK)
10.2.2: https://travis-ci.org/perl5-dbi/DBD-mysql/jobs/193932893 (FAIL)
10.2.3: https://travis-ci.org/perl5-dbi/DBD-mysql/jobs/193932894 (FAIL)

mbeijen:
I created https://jira.mariadb.org/browse/MDEV-11872 for this.

mbeijen:
So I merged #87. I think it's best to make a new PR with the `.travis.yml` and `Makefile.PL` changes from *this* PR, don't you think?

pali:
I squashed commits and rebased on top of master branch.

Now this pull request is ready for review. Part of this pull request is now also checking if compiler can found & use mysqlclient library via external module Devel::CheckLib. Fixes bug: https://rt.cpan.org/Ticket/Display.html?id=119902

pali:
So for now MariaDB 10.2.2 and newer stay disabled and not used by Travis until upstream fix that problem.

mbeijen:
MariaDB fix is committed here: https://github.com/MariaDB/server/commit/8877f1c32566b32bf9c598052e8f866a361190fb

pali:
Great! Seems it is working! And new MariaDB 10.2.4 with this fix was released, so we can add it into Travis matrix.

PR #87, opened 2017-01-21
Fix unit tests for different MySQL and MariaDB versions

pali:
Tests were fixed to correctly check for server or client versions (what is needed). Function MinimumVersion or attribute SQL_DBMS_VER do not work correctly as those returns strings which include also database server name (e.g. MariaDB vs MySQL).

Some tests are completely skipped if they depends on features not available in current tested MySQL server. And some tests were fixes to work with older and also newer MySQL and MariaDB versions as some function names was deprecated/renamed/removed.

With these changes all unit tests tested against MySQL and MariaDB versions in https://github.com/perl5-dbi/DBD-mysql/pull/85 are passing.

### DIFF
--- a/t/29warnings.t
+++ b/t/29warnings.t
@@ -17,8 +17,8 @@ if ($@) {
     plan skip_all => "no database connection";
 }
 
-if ( !MinimumVersion($dbh, '4.1') ) {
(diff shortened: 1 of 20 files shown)


mbeijen:
👍 Great work!

PR #88, opened 2017-01-28
Implement select() based implementation of async mode

pali:
select() is supported also by Windows and async mode is now enabled on them too.


pali:
Windows has also WSAPoll() function but it has bug 309411. The recommendation for now is to not use the WSAPoll function.

pali:
(hm... tests failing, do not merge for now)

pali:
The whole mysql_db_async_ready() is broken itself because perl header files redefine socket API.

pali:
I'm moving native socket code into separate C file which will not include any perl header file. This should fix all problems related to mysql sockets & redefined perl socket API.

pali:
Note that if select() code stay in dbdimp.c then select() would be redefined by perl.h to PerlSock_select() which would call my_perl->ISock->pSelect() and this perl implementation is incompatible with native one used by mysqlclient. And that call just returns error 10038 which means `An operation was attempted on something that is not a socket.`.

mbeijen:
Can you maybe rebase these commits into one?

pali:
Do you want to have all changes in one commit? I can do that...

I just separated move of socket code to separate file in first commit. In second implemented select(). And in third removed MYSQL_ASYNC check which is not needed...

mbeijen:
OK, if you prefer three separate commits I'll merge those.

pali:
Hm... something strange happened. My last commit 1638135626da25c3deaad51547d23908182ac99e passed checks, but your merge commit 071198b752515fec2166fb2d7657cee4b9ade7d1 (which is just empty diff!) did not pass...

pali:
There is difference in build log between those two checks:
https://ci.appveyor.com/project/mbeijen/dbd-mysql-mf7me/build/1.0.312 - StrawberryPerl v5.20.1.1 [Approved] - Possibly broken
https://ci.appveyor.com/project/mbeijen/dbd-mysql-mf7me/build/1.0.314 - StrawberryPerl v5.24.1.1 [Approved]

Looks like StrawberryPerl was upgraded on AppVeyor and new version does not provide libraries anymore... Or has some bugs :-(

mbeijen:
That's just great! :-(

The new Strawberry Perl provides MySQL 5.7.something libraries where this
used to be 5.1.

PR #91, opened 2017-01-29
Check only for libs via Devel::CheckLib in Makefile.PL

pali:
Problems with missing header files or cflags are already detected at compile time.


pali:
(do not merge until all test passes)

pali:
Seems it really fixes new Strawberry Perl on AppVeyor.

pali:
"All checks have passed"... so fixed.

PR #93, opened 2017-01-29
Run AppVeyor tests with different perl versions

pali:


pali:
So just two version 5.20.1.1 and 5.24.1.1 are working. Others available on https://chocolatey.org/packages/StrawberryPerl cannot be installed because cinst is trying to download them from non-working site http://strawberry-perl.googlecode.com/files/.

PR #94, opened 2017-02-02
Run AppVeyor tests with different perl versions and platforms

pali:
Now Chocolatey contains more Perl versions for both 32bit and 64bit
architectures. AppVeyor currently run on 64bit machine (also for 32bit
builds) and just set PLATFORM variable which needs to be correctly
propagated to cinst.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -2,14 +2,23 @@ version: 1.0.{build}
 
 environment:
   matrix:
-    - perl: "5.20.1.1"
+    - perl: "5.14.4.1"
+    - perl: "5.16.3.20170202"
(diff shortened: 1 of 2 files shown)


pali:
We have blinking test:
```
#   Failed test at t/87async.t line 142.
# Looks like you failed 1 test of 92.
t/87async.t ............................. 
Dubious, test returned 1 (wstat 256, 0x100)
Failed 1/92 subtests 
```

It randomly pass and randomly fails on Windows.

I force-pushed no changes to github and that test every time failed on different perl version...

pali:
Looks like that tests passed now.

PR #95, opened 2017-02-05
Fix mysql_auto_reconnect=1 when mysql_server_prepare=1 is used

pali:
Auto reconnect have not worked when DBD tried to prepare server side
statement and when to try execute server side prepared statement.


pali:
Seems I triggered some bug in MariaDB 10.1.2 which cause server crash! :-(

mbeijen:
... so is this ready for merging now?

pali:
No idea why first time that test did not passed... I just changed mirror for downloading MariaDB to main location and after that it passed.

So I think it is ready!

PR #98, opened 2017-02-05
Test MySQL & MariaDB Connector/C on Travis

pali:


pali:
(testing... do not merge)

pali:
That is really really bad :-( All Connector/C had broken or missing mysql_config script.

- MySQL Connector/C 6.0 - mysql_config has hardcoded path to `/usr/local/lib` and `/usr/local/include` and do not use relative paths like regular MySQL tarball 
- MySQL Connector/C 6.1 - mysql_config has nonsense libs flag set to `-l` without library name after that l -- that is invalid even for gcc
- MySQL Connector/C 6.2 - does not provide any mysql_config script
- MariaDB Connector/C (all versions) - does not provide any mysql_config script. There is only mariadb_config (ELF binary) which has hardcoded path to `/usr/local/include/mariadb` and `/usr/local/lib/mariadb`

And every Connector/C version has different name of client library. Some has libmysql.so, some libmysqlclient.so, some liblmysqlconc62.so and some libmariadb.so.

I tried to fix nonsence `-l` and generate some mysql_config script but it does not work for every different version.

So how to solve this problem? Correct should be to report it to MySQL & MariaDB projects to fix mysql_config scripts to provide correct compile flags and work correctly with relative paths (like in whole MySQL server package). But old versions stay broken...

pali:
MySQL Connector/C 6.0, 6.1 and 6.2 are now hacked/fixed. Either by fixing mysql_config or totally removing it and specifying flags manually.

With MariaDB Connector/C 2.2.0 there are two failed tests: t/40server_prepare_crash.t and t/rt118977-zerofill.t. With other 2.2 and 2.3 versions there is just one failed test t/rt118977-zerofill.t.

And MariaDB Connector/C 3.0.1 is totally incompatible with DBD::mysql. It has changed internal `MYSQL` structure and does not provide all mysql header files anymore...

pali:
Test t/rt118977-zerofill.t is failing with all versions of MariaDB Connector/C because of bug in MariaDB Connector/C library. That library ignores `ZEROFILL_FLAG` and so output is always returned without leading zeros, like without `zerofill` attribute.

Because of this bug all versions of MariaDB Connector/C does not work with DBD::mysql.

pali:
@mbeijen Can you report MariaDB & MySQL bugs from https://github.com/perl5-dbi/DBD-mysql/pull/98#issuecomment-277660862 and https://github.com/perl5-dbi/DBD-mysql/pull/98#issuecomment-278932984

pali:
Now all MySQL Connector/C versions are hacked to work on Travis. And MariaDB Connector/C are disabled due to zerofill bug. So patches are now ready.

pali:
@mbeijen PING

PR #99, opened 2017-02-07
Bail out tests when env CONNECTION_TESTING is 1 and connection to MySQL server is not possible

pali:
On AppVeyor and Travis CI is non-working MySQL or MariaDB server fatal
error which should not be ignored. This patch introduce new test function
DbiTestConnect() which is wrapper around DBI->connect() and based on env
CONNECTION_TESTING value either skip test or bail out. All tests are now
adjusted to use DbiTestConnect() on startup.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -102,6 +102,7 @@ install:
       chmod +x "$SANDBOX_HOME/msb/mysql_config" || exit 1 ;
     fi
   - export RELEASE_TESTING=1
+  - export CONNECTION_TESTING=1
 script:
   - if [ -n "$DB" ]; then
       perl Makefile.PL --mysql_config="$SANDBOX_HOME/msb/mysql_config" --testuser=msandbox --testpassword=msandbox --testhost=127.0.0.1 --testport=3310 && make disttest || exit 1;
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -31,4 +31,5 @@ build_script:
   - perl Makefile.PL --mysql_config=c:\strawberry\c\bin\mysql_config.bat --testuser=root --testpassword=Password12!
 
 test_script:
+  - set CONNECTION_TESTING=1
   - dmake test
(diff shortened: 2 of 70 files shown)


mbeijen:
I understand why you want this, but I'm not sure I like the solution. I think it's nicer to have some generic test that is run ONCE instead of modifying ALL tests.

pali:
Problem is that some tests and only sometime (!) caused server to not accept new connections (maybe server crashed or some other problem happened). And then next tests were skipped with successful result. So testing connection in one test at begining is not enough. If you have better solution let me know...

mbeijen:
I'll think about it!

pali:
Note that this new Bail out behavior is enabled only if CONNECTION_TESTING is set to 1 and that is enabled only for Travis and AppVeyor tests.

pali:
Here is example of this problem 
https://travis-ci.org/perl5-dbi/DBD-mysql/jobs/204914808:
```
t/15reconnect.t ......................... 1/26 DBD::mysql::db prepare failed: MySQL server has gone away at t/15reconnect.t line 46.
DBD::mysql::db prepare failed: MySQL server has gone away at t/15reconnect.t line 46.
# Looks like you planned 26 tests but ran 23.
# Looks like your test exited with 2 just after 23.
t/15reconnect.t ......................... Dubious, test returned 2 (wstat 512, 0x200)
Failed 3/26 subtests 
t/16dbi-get_info.t ...................... skipped: no database connection
t/20createdrop.t ........................ skipped: no database connection
t/25lockunlock.t ........................ skipped: no database connection
t/29warnings.t .......................... skipped: no database connection
t/30insertfetch.t ....................... skipped: no database connection
t/31insertid.t .......................... skipped: no database connection
t/32insert_error.t ...................... skipped: no database connection
t/35limit.t ............................. skipped: no database connection
t/35prepare.t ........................... skipped: no database connection
t/40bindparam.t ......................... ok
```

PR #105, opened 2017-02-26
Speed up builds on AppVeyor by caching downloaded files between builds

pali:
AppVeyor supports caching directories between builds. Enable caching of
StrawberryPerl installer and sources of cpan distributions. This should
decrease build time as AppVeyor will use cached files instead downloading
them from internet again and again in every build.


pali:
(testing, do not merge)

pali:
Seems that it is not possible to update cache on AppVeyor from pull requests builds. So I cannot do real time measurements...

Anyway, pull request is now ready for review.

pali:
Looks good, after merge cache is enabled and build time decreased to 27m which is lowest value...

PR #108, opened 2017-03-15
Set ssl_mode for Oracle MySQL if applicable.

dveeden:
Supersedes #82

### DIFF
--- a/dbdimp.c
+++ b/dbdimp.c
@@ -2052,16 +2052,14 @@ MYSQL *mysql_dr_connect(
 	    char *ca_path = NULL;
 	    char *cipher = NULL;
 	    STRLEN lna;
-#if MYSQL_VERSION_ID >= SSL_VERIFY_VERSION && MYSQL_VERSION_ID <= SSL_LAST_VERIFY_VERSION
-            /*
-              New code to utilise MySQLs new feature that verifies that the
-              server's hostname that the client connects to matches that of
-              the certificate
-            */
+#ifdef MYSQL_SSL_MODE
+	    unsigned int ssl_mode = SSL_MODE_PREFERRED;
+#endif
+            /* Verify if the hostname we connect to matches the hostname in the certificate */
 	    my_bool ssl_verify_true = 0;
 	    if ((svp = hv_fetch(hv, "mysql_ssl_verify_server_cert", 28, FALSE))  &&  *svp)
 	      ssl_verify_true = SvTRUE(*svp);
-#endif
+
 	    if ((svp = hv_fetch(hv, "mysql_ssl_client_key", 20, FALSE)) && *svp)
 	      client_key = SvPV(*svp, lna);
 
@@ -2083,10 +2081,32 @@ MYSQL *mysql_dr_connect(
 
 	    mysql_ssl_set(sock, client_key, client_cert, ca_file,
 			  ca_path, cipher);
-#if MYSQL_VERSION_ID >= SSL_VERIFY_VERSION && MYSQL_VERSION_ID <= SSL_LAST_VERIFY_VERSION
+#ifdef MYSQL_SSL_MODE
+	    if (ssl_verify_true)
+	      ssl_mode = SSL_MODE_VERIFY_IDENTITY;
+	    else if (ca_file)
+	      ssl_mode = SSL_MODE_VERIFY_CA;
+	    else if (ca_path)
+	      ssl_mode = SSL_MODE_VERIFY_CA;
+	    mysql_options(sock, MYSQL_OPT_SSL_MODE, &ssl_mode);
+#elif MYSQL_VERSION_ID >= SSL_VERIFY_VERSION && MYSQL_VERSION_ID <= SSL_LAST_VERIFY_VERSION || MYSQL_VERSION_ID >= MARIADB_VERSION_10
 	    mysql_options(sock, MYSQL_OPT_SSL_VERIFY_SERVER_CERT, &ssl_verify_true);
+#if MYSQL_VERSION_ID >= SSL_ENFORCE_VERSION && MYSQL_VERSION_ID <= SSL_LAST_ENFORCE_VERSION
+	    /* Only needed for Oracle MySQL 5.7 if MYSQL_OPT_SSL_MODE is not available */
+	    mysql_options(sock, MYSQL_OPT_SSL_ENFORCE, &ssl_verify_true);
+#endif
+#else
+	    croak("Can't enable strict certificate checks");
 #endif
 	    client_flag |= CLIENT_SSL;
+#ifdef MYSQL_SSL_MODE
+	  }
+	    else
+	  {
+	    /* mysql_ssl=0 */
+	    unsigned int ssl_mode = SSL_MODE_DISABLED;
+	    mysql_options(sock, MYSQL_OPT_SSL_MODE, &ssl_mode);
+#endif
 	  }
 	}
 #endif
--- a/dbdimp.h
+++ b/dbdimp.h
@@ -61,12 +61,17 @@
 #define NEW_DATATYPE_VERSION 50003
 #define SSL_VERIFY_VERSION 50023
 #define SSL_LAST_VERIFY_VERSION 50799
+#define SSL_ENFORCE_VERSION 50703
+#define SSL_LAST_ENFORCE_VERSION 50799
 #define MYSQL_VERSION_5_0 50001
+#define MARIADB_VERSION_10 100000
 /* This is to avoid the ugly #ifdef mess in dbdimp.c */
 #if MYSQL_VERSION_ID < SQL_STATE_VERSION
 #define mysql_sqlstate(svsock) (NULL)
 #endif
-
+#if MYSQL_VERSION_ID > 50710 && MYSQL_VERSION_ID < MARIADB_VERSION_10
+#define MYSQL_SSL_MODE
+#endif
 /*
  * This is the versions of libmysql that supports MySQL Fabric.
 */


pali:
With some versions of mysql client compilation failed... Looks like these continuous-integration tests are really good for catching problems with code which has lot of #ifdefs...

pali:
Anyway, I do not think that usage of SSL/TLS tunel without verifycation of server certificate (either by CA or certificate chain or directly by fingerprint of server certificate) is a good idea. In some cases it could make sense (e.g. passive scanning), but basically it has similar security as not using SSL/TLS tunel.

Without verification of server certificate MITM is possible in same way as MITM for plain text without SSL/TLS.

I really would suggest that verification of server certificate should be preferred option. We probably want to have backward compatibility, but at least examples and documentation can be updated to show people default usage of certificate verification.

In every case it is better if users of DBD::mysql starts doing certificate verification. And documentation or "example" usage could help...

dveeden:
> Anyway, I do not think that usage of SSL/TLS tunel without verifycation of server certificate (either by CA or certificate chain or directly by fingerprint of server certificate) is a good idea. In some cases it could make sense (e.g. passive scanning), but basically it has similar security as not using SSL/TLS tunel.
It is slightly better than not using TLS as it protects against passive attacks.
However it can also give a false sense of security and be a performance overhead.
So no clear winner there.

> Without verification of server certificate MITM is possible in same way as MITM for plain text without SSL/TLS.
Yes indeed.

> I really would suggest that verification of server certificate should be preferred option. We probably want to have backward compatibility, but at least examples and documentation can be updated to show people default usage of certificate verification.
Yes, VERIFY_CA is reasonably secure. VERIFY_IDENTITY is more secure, but hostname validation is in practice not always possible becuase MySQL and MariaDB don't support SubjectAlternativeName.

> In every case it is better if users of DBD::mysql starts doing certificate verification. And documentation or "example" usage could help...
Indeed.

dbdimp.c
```diff
@@ -2083,11 +2081,33 @@ MYSQL *mysql_dr_connect(
 
 	    mysql_ssl_set(sock, client_key, client_cert, ca_file,
 			  ca_path, cipher);
-#if MYSQL_VERSION_ID >= SSL_VERIFY_VERSION && MYSQL_VERSION_ID <= SSL_LAST_VERIFY_VERSION
+#ifdef MYSQL_SSL_MODE
+            if ssl_verify_true:
```
pali: syntax error. In C it is `if (ssl_verify_true)` and no colon at the end.

dbdimp.h
```diff
@@ -61,12 +61,17 @@
...
+#define SSL_ENFORCE_VERSION 50703
+#define SSL_LAST_ENFORCE_VERSION 50799
 #define MYSQL_VERSION_5_0 50001
+#define MARIADB_VERSION_10 100000
 /* This is to avoid the ugly #ifdef mess in dbdimp.c */
 #if MYSQL_VERSION_ID < SQL_STATE_VERSION
 #define mysql_sqlstate(svsock) (NULL)
 #endif
-
+#ifdef SSL_MODE_PREFERRED
```
pali: Looks like `SSL_MODE_PREFERRED` is not macro so this #ifdef is always false and `MYSQL_SSL_MODE` is never defined.

dbdimp.c
```diff
@@ -2083,11 +2081,33 @@ MYSQL *mysql_dr_connect(
 
 	    mysql_ssl_set(sock, client_key, client_cert, ca_file,
 			  ca_path, cipher);
-#if MYSQL_VERSION_ID >= SSL_VERIFY_VERSION && MYSQL_VERSION_ID <= SSL_LAST_VERIFY_VERSION
+#ifdef MYSQL_SSL_MODE
+            if (ssl_verify_true)
+              ssl_mode = SSL_MODE_VERIFY_IDENTITY;
+            else if (strlen(ca_file) > 0)
```
pali: ca_file and also ca_path could be NULL which means strlen here can crash

dbdimp.c
```diff
@@ -2083,11 +2081,33 @@ MYSQL *mysql_dr_connect(
...
+              ssl_mode = SSL_MODE_VERIFY_CA;
+	    mysql_options(sock, MYSQL_OPT_SSL_MODE, &ssl_mode);
+#elif MYSQL_VERSION_ID >= SSL_VERIFY_VERSION && MYSQL_VERSION_ID <= SSL_LAST_VERIFY_VERSION || MYSQL_VERSION_ID >= MARIADB_VERSION_10
 	    mysql_options(sock, MYSQL_OPT_SSL_VERIFY_SERVER_CERT, &ssl_verify_true);
+#if MYSQL_VERSION_ID >= SSL_ENFORCE_VERSION && MYSQL_VERSION_ID <= SSL_LAST_ENFORCE_VERSION
+            /* Only needed for Oracle MySQL 5.7 if MYSQL_OPT_SSL_MODE is not available */
+	    mysql_options(sock, MYSQL_OPT_SSL_ENFORCE, &ssl_verify_true);
+#endif
+#else
+            warn("Can't enable strict certificate checks");
```
pali: I would say this is a fatal error.

pali:
Now we have there Riddle vulnerability at http://riddle.link/

And looks like Oracle probably fixes SSL enforcement in 5.5 and 5.6 versions of MySQL as documentation was already updated. Compare:
https://web.archive.org/web/20161219180824/https://dev.mysql.com/doc/refman/5.5/en/mysql-ssl-set.html
https://dev.mysql.com/doc/refman/5.5/en/mysql-ssl-set.html

> In MySQL 5.5.55, the MYSQL_OPT_SSL_MODE option for mysql_options() was backported from MySQL 5.7 to MySQL 5.5.

But MySQL 5.5.55 was not released yet...

dbdimp.c
dveeden: Fixed

dbdimp.c
dveeden: Yes I agree. Note that currently there isn't even a warning in this case.

dbdimp.c
```diff
@@ -2083,11 +2081,33 @@ MYSQL *mysql_dr_connect(
...
+              ssl_mode = SSL_MODE_VERIFY_CA;
+	    mysql_options(sock, MYSQL_OPT_SSL_MODE, &ssl_mode);
+#elif MYSQL_VERSION_ID >= SSL_VERIFY_VERSION && MYSQL_VERSION_ID <= SSL_LAST_VERIFY_VERSION || MYSQL_VERSION_ID >= MARIADB_VERSION_10
 	    mysql_options(sock, MYSQL_OPT_SSL_VERIFY_SERVER_CERT, &ssl_verify_true);
+#if MYSQL_VERSION_ID >= SSL_ENFORCE_VERSION && MYSQL_VERSION_ID <= SSL_LAST_ENFORCE_VERSION
+            /* Only needed for Oracle MySQL 5.7 if MYSQL_OPT_SSL_MODE is not available */
+	    mysql_options(sock, MYSQL_OPT_SSL_ENFORCE, &ssl_verify_true);
+#endif
+#else
+            die("Can't enable strict certificate checks");
```
pali: Use `croak(...)` as you are not passing return value from `die(...)` somewhere else. Look into perlapi for details.

dbdimp.c
```diff
@@ -2083,11 +2081,33 @@ MYSQL *mysql_dr_connect(
...
 			  ca_path, cipher);
-#if MYSQL_VERSION_ID >= SSL_VERIFY_VERSION && MYSQL_VERSION_ID <= SSL_LAST_VERIFY_VERSION
+#ifdef MYSQL_SSL_MODE
+            if (ssl_verify_true)
+              ssl_mode = SSL_MODE_VERIFY_IDENTITY;
+            else if (ca_file && (strlen(ca_file) > 0))
+              ssl_mode = SSL_MODE_VERIFY_CA;
+            else if (ca_path && (strlen(ca_path) > 0))
+              ssl_mode = SSL_MODE_VERIFY_CA;
+	    mysql_options(sock, MYSQL_OPT_SSL_MODE, &ssl_mode);
```
pali: Seems your patch is whitespace inconsistence with code around... See indentation. If it is possible try to fix it...

dbdimp.c
```diff
@@ -2083,11 +2081,33 @@ MYSQL *mysql_dr_connect(
 
 	    mysql_ssl_set(sock, client_key, client_cert, ca_file,
 			  ca_path, cipher);
-#if MYSQL_VERSION_ID >= SSL_VERIFY_VERSION && MYSQL_VERSION_ID <= SSL_LAST_VERIFY_VERSION
+#ifdef MYSQL_SSL_MODE
+            if (ssl_verify_true)
+              ssl_mode = SSL_MODE_VERIFY_IDENTITY;
+            else if (ca_file && (strlen(ca_file) > 0))
```
pali: Do we need check for `strlen()`? I have not find any information about it and variables are already passed to `mysql_ssl_set()` without `strlen()` check... So I'm not sure there...

dveeden:
> This code starts to be a big #if mess with long conditions after #if... I do not have idea how to improve it right now.

I'm afraid this will get worse over time because MySQL and MariaDB are (very) slowly diverging. We can try to prevent that as much as possible, but eventually it might make sense to split this into DBD::mysql and DBD::mariadb.

pali:
> I'm afraid this will get worse over time because MySQL and MariaDB are (very) slowly diverging.

Probably...

> We can try to prevent that as much as possible, but eventually it might make sense to split this into DBD::mysql and DBD::mariadb.

But here we have another problem unrelated to MariaDB... Problem is with different MySQL versions. So splitting into DBD::mysql and DBD::mariadb does not help there. And there are still very large common codebase so I do not think split is good idea.

dbdimp.c
dveeden: Removed strlen.

dbdimp.c
dveeden: I've tried to fix it.

dbdimp.c
dveeden: I've fixed that

dveeden:
If `mysql_ssl=1` is not present in the dsn but something like `mysql_ssl_ca_file` is then the latter is ignored. I think we should generate a warning in that case. Not sure what the best way is to do that and I'd like to fix that in a separate pull request.

pali:
I did some SSL changes here: https://github.com/perl5-dbi/DBD-mysql/pull/114 I tried to make that ifdef hell more readable and prepare ground for new MySQL 5.5.55 which should finally have fixed Riddle vulnerability.

dveeden:
@pali should we wait on your ssl branch to be ready to merge? Or should this branch be merged?
I'd like to see something merged to fix the insecure code (not related to Riddle).

pali:
For me this is PR is ready to merge, I already marked it.

pali:
I updated my https://github.com/perl5-dbi/DBD-mysql/pull/114 on top of master (so including this PR).

pali:
@dveeden this PR #108 broke compilation for MySQL C/Connector 6.0:
https://travis-ci.org/perl5-dbi/DBD-mysql/builds/218257554

Issue #110, opened 2017-03-23
Handle SSL/TLS correctly


pali:
Currently DBD::mysql has these SSL/TLS connections parameters:
```
mysql_ssl
mysql_ssl_client_key
mysql_ssl_client_cert
mysql_ssl_ca_file
mysql_ssl_ca_path
mysql_ssl_cipher
mysql_ssl_verify_server_cert
```

After discussion in https://github.com/perl5-dbi/DBD-mysql/pull/82 and https://github.com/perl5-dbi/DBD-mysql/pull/108 I'm proposing following change how DBD::mysql should process SSL/TLS settings:

1. Parameter `mysql_ssl=1` would change its meaning and after this change SSL/TLS will be required and enforced. If server does not support it then client must not connect to MySQL server and must reject connection. `mysql_ssl=0` would mean that SSL/TLS is disabled (default value).

2. Introduce new `mysql_ssl_optional=1` parameter (default false; SSL/TLS is required and enforced) which could allow to connect to MySQL server without SSL/TLS support when `mysql_ssl=1`. This is dangerous as BACKRONYM (http://backronym.fail/) or Riddle (http://riddle.link/) can take effect, but in some cases it could be useful (e.g. when it can be ensured that it is not possible to modify any packet between client and server, just there can be passive monitoring of network). Documentation for `mysql_ssl_optional=1` must describe this problem and suggest users to not enable this option if they are unsure.

3. When specified `mysql_ssl_ca_file` or `mysql_ssl_ca_path` then DBD::mysql must check and verify CA certificate. If validation fails then connection must be rejected. Certificate hostname is not validated until `mysql_ssl_verify_server_cert=1` is set.

4. If DBD::mysql (or underlaying libmysqlclient.so library) decide to reject connection with MySQL server then data (login credentials, authentication negotiation or SQL statements) must not be sent to MySQL server.

5. If underlaying libmysqlclient.so library is not able to enforce current configuration specified by `mysql_ssl*` parameters then it must reject connection.

6. If it is possible try to support different versions of libmysqlclient.so library (MySQL, MariaDB, 5.5, 5.6, 5.7, 8.0, ...). If not, rather drop SSL support for particular libmysqlclient.so version as providing incorrect/broken/vulnerable SSL encryption.

[...]

PR #114, opened 2017-04-02
Improve SSL settings, reflect changes for BACKRONYM and Riddle vulnerabilities, enforce SSL encryption when mysql_ssl=1 is set

pali:
This pull request improve SSL settings by:
* Describing all SSL related attributes in POD documentation
* Reflect changes between different versions of libmysqlclient.so and properly enforce SSL encryption
* Fixed BACKRONYM and Riddle vulnerabilities
* Enforce SSL encryption when mysql_ssl=1 is set
* Add new connection attribute mysql_ssl_optional
* Add new database handle attribute mysql_ssl_cipher

The important change is that DBD::mysql reject connection to MySQL server (also SSL enabled) if mysql_ssl=1 is set and libmysqlclient.so library cannot enforce SSL encryption (because is vulnerable to BACKRONYM or Riddle).

See also discussion at https://github.com/perl5-dbi/DBD-mysql/issues/110

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,11 @@ matrix:
     - perl: "5.20"
       env: DB=MySQL VERSION=5.5.47
     - perl: "5.20"
+      env: DB=MySQL VERSION=5.5.49
+    - perl: "5.20"
       env: DB=MySQL VERSION=5.5.54
+    - perl: "5.20"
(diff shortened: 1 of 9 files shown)


agx:
This should be reopened now that master got reverted back to 4.041

PR #116, opened 2017-04-04
Fix build with Connector/C 6.0.x

dveeden:
MySQL Connector/C 6.0.x sets the MySQL version to 6.0.0
MySQL Connector/C 6.1.x sets the MySQL version to 5.7.x

Fixes the broken build as reported in #108 

This defines `MYSQL_SSL_MODE` for MySQL 5.7.11 and up including MySQL 8.0.0.
But excludes 6.0.0 as that's what Connector/C 6.0.x uses for some reason.


dveeden:
cc @pali

PR #124, opened 2017-05-01
Make mysql_async_ready() and mysql_async_result() idempotent

pali:
`$sth->mysql_async_ready` now returns 1 also after `$sth->mysql_async_result`. And duplicate `$sth->mysql_async_result` now returns same value (repeated time just cached value, instead throwing error).
    
This change would allow to call `$sth->mysql_async_result` more times without changing internal state or throwing errors.

And afterword it allow to call `$sth->mysql_async_ready` and `$sth->mysql_async_result` before all fetch* $sth methods which fixes bugs:

https://rt.cpan.org/Public/Bug/Display.html?id=103306
https://github.com/jhthorsen/mojo-mysql/issues/1

Note that it changes behavior of both `mysql_async_ready` and `mysql_async_result` methods. Before this change both methods returned `undef` if corresponding handle was not currently running an asynchronous query. After this change both methods return `undef` if corresponding handle was not created with 'async' set to true or if an asynchronous query was not started yet.

### DIFF
--- a/dbdimp.c
+++ b/dbdimp.c
@@ -5896,6 +5896,7 @@ int mysql_db_async_result(SV* h, MYSQL_RES** resp)
   MYSQL_RES* _res;
   int retval = 0;
   int htype;
+  bool async_sth = FALSE;
 
   if(! resp) {
       resp = &_res;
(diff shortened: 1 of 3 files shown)


pali:
looks like a travis problem...
```
dpkg: error: dpkg status database is locked by another process
apt-get install failed
```

pali:
Test failed due to some travis related problem with apt...

PR #127, opened 2017-05-23
Added skip logic for unsupported MySQL Server versions

ikari7789:
Added logic to skip tests when they are ran against a version of MySQL that doesn't support the feature being tested.

### DIFF
--- a/t/55utf8_jp.t
+++ b/t/55utf8_jp.t
@@ -11,6 +11,11 @@ require "lib.pl";
 
 my $dbh = DbiTestConnect($test_dsn, $test_user, $test_password, { mysql_enable_utf8 => 1, PrintError => 1, RaiseError => 1 });
 
+if (!MinimumVersion($dbh, '5.5')) {
+    plan skip_all =>
+        "SKIP TEST: You must have MySQL version 5.5 or greater for this test to run";
+}
+
 eval {
   $dbh->{PrintError} = 0;
   $dbh->do("SET lc_messages = 'ja_JP'");
--- a/t/cve-2017-3302.t
+++ b/t/cve-2017-3302.t
@@ -1,4 +1,10 @@
 # This must be minimal test, even strict or Test::More can hide real crash
+# This test should be restricted to affected versions:
+#   MySQL   <= 5.7.5
+#   MariaDB <= 5.5.54
+#   MariaDB <= 10.0.29
+#   MariaDB <= 10.1.21
+#   MariaDB <= 10.2.3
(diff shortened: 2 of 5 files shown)


mbeijen:
Those are OLD perl versions! 👍

t/cve-2017-3302.t
```diff
@@ -9,7 +15,7 @@ use vars qw($test_dsn $test_user $test_password $test_db);
 use lib 't', '.';
 require "lib.pl";
 eval {
-  $dbh = DBI->connect($test_dsn, $test_user, $test_password, {RaiseError => 1, mysql_server_prepare => 1});
```
pali: Hey!! What are you doing? Whole test needs to enable by default prepared statement as it tests CVE-2017-3302. Please revert whole changes in this file. It should test on every version, not just some affected!

t/55utf8_jp.t
```diff
@@ -11,6 +11,11 @@ require "lib.pl";
 
 my $dbh = DbiTestConnect($test_dsn, $test_user, $test_password, { mysql_enable_utf8 => 1, PrintError => 1, RaiseError => 1 });
 
+if (!MinimumVersion($dbh, '5.5')) {
+    plan skip_all =>
+        "SKIP TEST: You must have MySQL version 5.5 or greater for this test to run";
```
pali: Why is this test skipped before 5.5? On Travis it is working...
MySQL 5.1: https://travis-ci.org/perl5-dbi/DBD-mysql/jobs/227161603
MySQL 5.0: https://travis-ci.org/perl5-dbi/DBD-mysql/jobs/227161601
MySQL 4.1: https://travis-ci.org/perl5-dbi/DBD-mysql/jobs/227161599

ikari7789: `lc_messages` was added from MySQL version 5.5 as far as I could tell from the reference manuals. If it's available in earlier versions then it's undocumented behavior in my opinion.

t/cve-2017-3302.t
ikari7789: Oh, I just realized that this CVE is related to prepared statements. I'll re-add the check. I was trying to keep the test minimal as stated as it will fail on MySQL 4.0.

t/55utf8_jp.t
pali: But we have already check if `lc_messages` messages are available. This test uses `lc_messages` and not particular MySQL version. So test plan should check for existence of `lc_messages` and not MySQL version if it is possible. And this test is there.

Issue #126, opened 2017-05-22
Use ppport.h for adding backwards-compatibility


ikari7789:
DBD-mysql since v4.041 cannot compile on Perl <~ 5.8.9 due to the addition of PERL_UNUSED_ARG. This should be able to be remedied by using `ppport.h` but I'm not familiar with XS code at all, so I can't get it working to make a pull request.

I was able to get 4.042 to compile on my machine by adding the following to the top of dbdimp.c, but it doesn't feel like the appropriate solution.

```c
/* copied from ppport.h 3.36 */
#ifndef PERL_UNUSED_ARG
#  if defined(lint) && defined(S_SPLINT_S) /* www.splint.org */
#    include <note.h>
#    define PERL_UNUSED_ARG(x) NOTE(ARGUNUSED(x))
#  else
#    define PERL_UNUSED_ARG(x) ((void)x)
#  endif
#endif
```

Tux:
In *theory*, it should be
```
$ cpan Devel::PPPort
$ perl -MDevel::PPPort -we'Devel::PPPort::WriteFile ("ppport.h")'
$ perl ppport.h --compat-version=5.8.1 --patch=dp.patch
$ patch -p0 <dp.patch
```
But when I then start the regular build process, many things blow up. I have no time to dig

pali:
Here is pull request for Perl 5.8.1 support: https://github.com/perl5-dbi/DBD-mysql/pull/130 We do not need ppport. Pull request also enable Travis CI tests with Perl 5.8.1 version.

ikari7789:
Thanks @pali!

PR #129, opened 2017-05-23
Added in a shim for PERL_UNUSED_ARG in order to compile on Perl 5.8.1

ikari7789:
This should be replaced with ppport.h, but I’m unable to get it working correctly.


ikari7789:
Adds a fix for the problem mentioned in #126, but doesn't solve the problem entirely.

PR #131, opened 2017-05-24
Revert test edits

ikari7789:
Reverted the edits made in `t/cve-2017-3302.t` as per @pali's request.

Changed logic to check minimum server version? @pali not sure if that is OK or not. If not, I'll remove the extra check.

Fixed if logic in `t/10connect.t` as it checked version number in the wrong order.

### DIFF
--- a/t/cve-2017-3302.t
+++ b/t/cve-2017-3302.t
@@ -10,9 +10,10 @@ use lib 't', '.';
 require "lib.pl";
 eval {
   $dbh = DBI->connect($test_dsn, $test_user, $test_password, {RaiseError => 1, mysql_server_prepare => 1});
+  die 'MySQL version less than 4.1 does not support prepared statements' unless MinimumVersion($dbh, '4.1');
 } or do {
   $@ = "unknown error" unless $@;
-  $@ =~ s/ at \S+ line \d+\s*$//;
+  $@ =~ s/ at \S+ line \d+\.?\s*$//;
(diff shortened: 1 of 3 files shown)


mbeijen:
Hi @pali - is this better?

t/cve-2017-3302.t
```diff
@@ -15,10 +9,11 @@ use vars qw($test_dsn $test_user $test_password $test_db);
 use lib 't', '.';
 require "lib.pl";
 eval {
-  $dbh = DBI->connect($test_dsn, $test_user, $test_password, {RaiseError => 1});
+  $dbh = DBI->connect($test_dsn, $test_user, $test_password, {RaiseError => 1, mysql_server_prepare => 1});
+  die 'MySQL version less than 4.1 does not support prepared statements' unless MinimumVersion($dbh, '4.1');
```
pali: This is incorrect. You need both client and server at version 4.1. Not just server.

ikari7789: Is it necessary though? I'm not super familiar with C, but it seems to me that version checks are already in place at compile time for the client versions. If the client doesn't support prepared statements, then the attributes aren't even applied.

https://github.com/perl5-dbi/DBD-mysql/blob/master/dbdimp.c#L2039-L2067

ikari7789: Anyway, I added the client check.

pali: Yes, because prepared statements depends on both client and server. If code is not compiled, then test file would test something different.

PR #132, opened 2017-05-24
Fix detection of INNODB engine in tests

pali:
When connecting to MariaDB server 10.1.x via MySQL client 5.5 then MySQL
client get information that server version is 5.5.5.

Variable have_innodb was removed in MySQL server 5.6 and is also not
available in MariaDB server 10.1. So INNODB engine is not detected.

So when possible use SHOW ENGINES which is supported since MySQL 4.1.2.

### DIFF
--- a/t/rt75353-innodb-lock-timeout.t
+++ b/t/rt75353-innodb-lock-timeout.t
@@ -18,7 +18,7 @@ if (!@ilwtenabled) {
 }
 
 my $have_innodb = 0;
-if (!MinimumVersion($dbh1, '5.6')) {
+if (!MinimumVersion($dbh1, '4.1.2')) {
   my $dummy;
   ($dummy,$have_innodb)=
     $dbh1->selectrow_array("SHOW VARIABLES LIKE 'have_innodb'")
--- a/t/40catalog.t
+++ b/t/40catalog.t
@@ -31,7 +31,7 @@ SKIP: {
     if !MinimumVersion($dbh, '5.0');
 
   my $have_innodb = 0;
-  if (!MinimumVersion($dbh, '5.6')) {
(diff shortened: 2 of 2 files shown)


ikari7789:
MySQL 4.2.1 doesn't exist. Shouldn't it be 4.1.2?

pali:
Yes, fixed.

t/rt75353-innodb-lock-timeout.t
```diff
@@ -18,7 +18,7 @@ if (!@ilwtenabled) {
 }
 
 my $have_innodb = 0;
-if (!MinimumVersion($dbh1, '5.6')) {
+if (!MinimumVersion($dbh1, '4.1.2')) {
```
ikari7789: Minimum version for `have_innodb` is 3.23.37 according to the reference manual.

```
have_innodb

YES if mysqld supports InnoDB tables. DISABLED if --skip-innodb is used. This variable was
added in MySQL 3.23.37.
```

pali: Above test is for usage of SHOW ENGINES, have_innodb is fallback.

And I personally do not think that it is possible to use current DBD::mysql with older libmysqlclient as 4.1. Our Travis CI tests cover MySQL 4.1 as the oldest version.

ikari7789:
It is possible, and I am using it (and testing it) against a MySQL 4.0 server. It doesn't hurt to lower the minimum version to the one where it is available per the documentation for the test. It isn't about libmysqlclient but rather the server supporting it. I can look into creating a MySQL 4.0 tarball for the Travis tests if you'd like.

pali:
If support for MySQL 4.0 is really needed, then you can try to integrate it into Travis CI. CI would allow to automatically detect problems and also show us result if tests really passes.

PR #133, opened 2017-05-28
Fix for builing DBD-mysql together with MariaDB Connector/C.

9EOR9:
Use mysql_option function instead of accessing internal members of MYSQL structure.

### DIFF
--- a/dbdimp.c
+++ b/dbdimp.c
@@ -2252,6 +2252,7 @@ MYSQL *mysql_dr_connect(
 
     if (result)
     {
+      my_bool reconnect= 0;
 #if MYSQL_VERSION_ID >=SERVER_PREPARE_VERSION
       /* connection succeeded. */
       /* imp_dbh == NULL when mysql_dr_connect() is called from mysql.xs
@@ -2268,7 +2269,7 @@ MYSQL *mysql_dr_connect(
         we turn off Mysql's auto reconnect and handle re-connecting ourselves
         so that we can keep track of when this happens.
       */
-      result->reconnect=0;
+      mysql_options(result, MYSQL_OPT_RECONNECT, &reconnect);
     }
     else {
       /* 
--- a/dbdimp.c
+++ b/dbdimp.c
@@ -1961,7 +1961,7 @@ MYSQL *mysql_dr_connect(
           mysql_options(sock, MYSQL_READ_DEFAULT_GROUP, gr);
         }
         /* 60000 is identifier for MySQL Connector/C versions 6.0.x which do not support MYSQL_OPT_CONNECT_ATTR_ADD */
-        #if (MYSQL_VERSION_ID >= 50606 && MYSQL_VERSION_ID != 60000)
+        #ifdef CLIENT_CONNECT_ATTRS
           if ((svp = hv_fetch(hv, "mysql_conn_attrs", 16, FALSE)) && *svp) {
               HV* attrs = (HV*) SvRV(*svp);
               HE* entry = NULL;
@@ -4235,7 +4235,10 @@ int dbd_describe(SV* sth, imp_sth_t* imp_sth)
       default:
         if (buffer->buffer_type != MYSQL_TYPE_BLOB)
           buffer->buffer_type= MYSQL_TYPE_STRING;
-        buffer->buffer_length= fields[i].max_length ? fields[i].max_length : 1;
+        /* In case we didn't call mysql_stmt_attr_set with option
+           STMT_ATTR_UPDATE_MAX_LENGTH, we need to recalculate length */
+        buffer->buffer_length= fields[i].max_length ? fields[i].max_length :
+                               fields[i].length ? fields[i].length : 1;
         Newz(908, fbh->data, buffer->buffer_length, char);
         buffer->buffer= (char *) fbh->data;
         break;
--- a/dbdimp.h
+++ b/dbdimp.h
@@ -19,10 +19,14 @@
  */
 #include <DBIXS.h>  /* installed by the DBI module                        */
 #include <mysql.h>  /* Comes with MySQL-devel */
-#include <mysqld_error.h>  /* Comes MySQL */
 
 #include <errmsg.h> /* Comes with MySQL-devel */
 #include <stdint.h> /* For uint32_t */
+
+/* MariaDB Connector/C doesn't ship GPL'ed licensed mysqld_error.h
+   so we use hardcoded error code here instead of including mysqld_error.h
+ */
+#define ER_UNSUPPORTED_PS 1295
 
 #ifndef PERL_STATIC_INLINE
 #define PERL_STATIC_INLINE static
@@ -60,7 +64,7 @@
 
 
 /*
- * This is the version of MySQL wherer
+ * This is the version of MySQL where
  * the server will be used to process prepare
  * statements as opposed to emulation in the driver
 */
@@ -75,6 +79,12 @@
 #define NEW_DATATYPE_VERSION 50003
 #define MYSQL_VERSION_5_0 50001
 /* This is to avoid the ugly #ifdef mess in dbdimp.c */
+
+/* MariaDB Connector/C has MARIADB_VERSION_ID instead of MYSQL_VERSION_ID */
+#if !defined(MYSQL_VERSION_ID) && defined(MARIADB_VERSION_ID)
+#define MYSQL_VERSION_ID MARIADB_VERSION_ID
+#endif
+
 #if MYSQL_VERSION_ID < SQL_STATE_VERSION
 #define mysql_sqlstate(svsock) (NULL)
 #endif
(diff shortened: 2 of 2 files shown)


pali:
Some tests failed:
```
dbdimp.c:2272:29: error: ‘MYSQL_OPT_RECONNECT’ undeclared (first use in this function)
```

Also note that MariaDB Connector/C does not work with DBD::mysql, because MariaDB Connector/C has broken zerofill support, see: https://github.com/perl5-dbi/DBD-mysql/commit/e1ea5521675759634ad9e2e83118fd721d679a26 https://github.com/perl5-dbi/DBD-mysql/pull/98#issuecomment-278932984

pali:
This could help you:

http://web.archive.org/web/20110902122038/http://dev.mysql.com/doc/refman/5.0/en/auto-reconnect.html
> The MYSQL_OPT_RECONNECT option is available as of MySQL 5.0.13.

dbdimp.h
```diff
@@ -110,22 +124,22 @@
  */
 
 /* Use mysql_options with MYSQL_OPT_SSL_VERIFY_SERVER_CERT */
-#if ((MYSQL_VERSION_ID >= 50023 && MYSQL_VERSION_ID < 50100) || MYSQL_VERSION_ID >= 50111) && (MYSQL_VERSION_ID < 80000 || defined(MARIADB_BASE_VERSION))
+#ifdef CLIENT_SSL_VERIFY_SERVER_CERT
```
9EOR9: The verification is still in place:
https://github.com/mysql/mysql-server/blob/8.0/sql-common/client.cc#L3715

Even if MYSQL_OPT_SSL_VERIFY_SERVER_CERT was renoved, you can still pass CLIENT_SSL_VERIFY_SERVER_CERT flag to mysql_real_connect (last parameter).

9EOR9: The verification is still in place:
https://github.com/mysql/mysql-server/blob/8.0/sql-common/client.cc#L3715

Even if `MYSQL_OPT_SSL_VERIFY_SERVER_CERT` was renoved, you can still pass CLIENT_SSL_VERIFY_SERVER_CERT flag to mysql_real_connect (last parameter).

pali: But this check is for defining `HAVE_SSL_VERIFY`. And it means DBD::mysql can use mysql_options() with `MYSQL_OPT_SSL_VERIFY_SERVER_CERT`. It has nothing with `CLIENT_SSL_VERIFY_SERVER_CERT` flag or mysql_real_connect() function.

Some of those mysql_* SSL related functions are broken in particular MySQL versions and therefore we need to correctly ifdef them and use safe variant.

See https://github.com/perl5-dbi/DBD-mysql/pull/114 and https://github.com/perl5-dbi/DBD-mysql/issues/110

Your change break SSL verification for some MySQL versions.

pali: So... I guess correct should be this check which do not break SSL verification:
```
#if ((MYSQL_VERSION_ID >= 50023 && MYSQL_VERSION_ID < 50100) || MYSQL_VERSION_ID >= 50111) && (MYSQL_VERSION_ID < 80000 || defined(MARIADB_CLIENT))
```

dbdimp.c
```diff
@@ -2268,7 +2269,7 @@ MYSQL *mysql_dr_connect(
         we turn off Mysql's auto reconnect and handle re-connecting ourselves
         so that we can keep track of when this happens.
       */
-      result->reconnect=0;
```
pali:
```
#if MYSQL_VERSION_ID >= 50013
mysql_options(result, MYSQL_OPT_RECONNECT, &reconnect);
#else
result->reconnect=0;
#endif
```

dbdimp.h
9EOR9: Yes, this looks better, I will fix and push later today. We need to think about to add ssl_mode also in Connector/C, so API's will not differ too much and iit would be easier to handle TLS/SSL options in the future (e.g. Connector/C additionally supports finger print check of server certificate).

AdamWill:
Note downstream (Fedora) bug: https://bugzilla.redhat.com/show_bug.cgi?id=1470196 , where one of our mariadb maintainers suggests adding `+#include <mysql_version.h>  /* For MYSQL_VERSION_ID */` for the versioning stuff...

pali:
@AdamWill that would not work, because `MYSQL_VERSION_ID` was added only recently in commit MariaDB Connector/C commit: https://github.com/MariaDB/mariadb-connector-c/commit/422d0f786e5e6eb103000a75738322c8baa9aa12

AdamWill:
@pali but no stable version was ever actually released without it, AFAICT. On the 2.x branch, nothing was renamed to `MARIADB`, so it still had `MYSQL_VERSION_ID`, see e.g. the 2.3.2 tag:

https://github.com/MariaDB/mariadb-connector-c/blob/v2.3.2/include/mysql_version.h.in

and in 3.0.2, the commit you mention was already included:

https://github.com/MariaDB/mariadb-connector-c/blob/v3.0.2/include/mariadb_version.h.in

So the only tagged release that does not have `MYSQL_VERSION_ID` is 3.0.1-beta, which was not a stable release. Does this still concern you?

pali:
Ou right, so problematic is only the 3.0.1-beta and all older versions are working... Then we can use it.

AdamWill:
Well, see my edit (the file still seems to have a different name on the 3.x branch). Also note the 'downstream bug' I referenced isn't actually for quite the same thing, it's about building against MariaDB 10.2 vs. 10.1 (not building against mariadb-connector-c).

Actually the more I look at this 3.x branch of mariadb-connector-c, the weirder it gets...AIUI, mariadb-connector-c / mysql-connector-c are supposed to be just the client library from the 'main' distribution (mariadb / mysql) isolated so you don't have to install the whole thing if you're just building a client. But mariadb-connector-c 3.x doesn't seem to match the client library from mariadb 10.2 at all, at least in terms of these header files. They haven't all been renamed in MariaDB itself - there's no `mariadb_` or `ma_` headers in https://github.com/MariaDB/server/tree/10.2/include . So that all seems a bit odd...

9EOR9:
Both MariaDB 10.2.7 (released last week) and Connector/C 3.0.2 (will be released today) have MYSQL_VERSION_ID

9EOR9:
Adam,
Connector/C 3.0 is identical to the client library in 10.2. Please note that Connector/C in server tree is a git submodule in directory libmariadb (Connector/C branch 10.2-server).

AdamWill:
@9EOR9 But what about the header file names? They seem to be different between the two...

9EOR9:
What header file names do you mean? The only file which should be included is mysql.h - it includes all necessary files for client applications.

AdamWill:
Hmm, I see. Well, the patch our maintainer suggested included an explicit `#include mysql_version.h` - so I guess that should be left out...

AdamWill:
@9EOR9 So poking into this further, I found what seems to be a bug in MariaDB itself: when you run an install of MariaDB 10.2, you ultimately wind up with the `mysql.h` from the `libmariadb/` subdirectory (which, as you pointed out, is mariadb-connector-c as a subproject) installed. *Not* the `mysql.h` from MariaDB itself. So you get the `mysql.h` for the client library, which ultimately doesn't provide as much stuff as the 'full fat' `mysql.h` for the whole of MariaDB. This ultimately seems to be behind several of the build failures we observed with things that want to use stuff that's defined in MariaDB's `mysql_version.h`, which MariaDB's `mysql.h` includes, but *not* in the client library's `mariadb_version.h`, which the client library's `mysql.h` includes. Reported as https://jira.mariadb.org/browse/MDEV-13370 .

GieltjE:
The bug has been fixed by MariaDB in 10.2.9, we really should be working on a fix as half the world is migrating to MariaDB and a lot of packages depend on this package (spamassassin, opendkim, opendmarc)

noplanman:
Any update on this? Has just smacked me in the face 😕

mannih:
FWIW: I just ran into this problem after upgrading my fedora install from 26 to 27. I was able to install DBD::mysql after applying this patch set: https://src.fedoraproject.org/rpms/perl-DBD-MySQL/blob/c9a8c6732a3cad7016fe0cbf217a9f0846af6bb9/f/DBD-mysql-4.043-Fix-build-failures-for-MariaDB-10.2.patch
Of course, I have no idea how well this works on other systems with other libraries.

GieltjE:
@mannih Just tested it on debian with MariaDB 10.2.12 with spamassassin, opendkim and opendmarc, everything working as it should (also applied the other pathes availble in https://src.fedoraproject.org/rpms/perl-DBD-MySQL/tree/master).

pali:
Slightly different approach was used to fix these problems in https://github.com/gooddata/DBD-MariaDB repository.

mbeijen:
Partly fixed with 0b1884f4153d3de0a91bae4d3b3cc1771ffdf4f3

dveeden:
As far as I'm aware this PR is not needed anymore. Please ping me if that is not the case, then we can re-open this PR and actually get the issue fixed.

PR #138, opened 2017-06-14
ChopBlanks should not trim binary fields

urcheon:
Binary fields are incorrectly being trimmed of trailing spaces when non server side prepare is being used.
Trimming of server side prepare binary fields is catered for properly.

"As of MySQL 5.0.3, trailing spaces are retained when values are stored and retrieved, in conformance with standard SQL. Before MySQL 5.0.3, trailing spaces are removed from values when they are stored into a VARCHAR column; this means that the spaces also are absent from retrieved values."

This means that testing VARCHAR (and VARBINARY as it turns out) on 4.1.22 will cause the test cases to fail.
However, since we're testing BLOB, TINYBLOB, MEDIUMBLOB and LONGBLOB, we have enough binary fields to confirm that after the change we do not trim the trailing spaces from binary fields after retrieval on the client side.

### DIFF
--- a/t/50chopblanks.t
+++ b/t/50chopblanks.t
@@ -14,7 +14,8 @@ eval {$dbh= DBI->connect($test_dsn, $test_user, $test_password,
 if ($@) {
     plan skip_all => "no database connection";
 }
-plan tests => 36 * 2;
+
+plan tests => (8 + ((5 + 9 + 9) * 4)) * 2;
 
 for my $mysql_server_prepare (0, 1) {
 eval {$dbh= DBI->connect("$test_dsn;mysql_server_prepare=$mysql_server_prepare;mysql_server_prepare_disable_fallback=1", $test_user, $test_password,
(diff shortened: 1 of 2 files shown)


t/50chopblanks.t
```diff
@@ -26,42 +26,59 @@ ok $dbh->do("DROP TABLE IF EXISTS dbd_mysql_t50chopblanks"), "drop table if exis
 my $create= <<EOT;
 CREATE TABLE dbd_mysql_t50chopblanks (
   id INT(4),
-  name TEXT
+  c_varchar VARCHAR(10),
```
pali: VARCHAR is problematic. MySQL server can configured to automatically chop blanks. We could expects that this test start randomly on cpan failing. See commit: ce14f6c989eb3e626586a24aed16ee6ebcf18a82

t/50chopblanks.t
```diff
@@ -26,42 +26,59 @@ ok $dbh->do("DROP TABLE IF EXISTS dbd_mysql_t50chopblanks"), "drop table if exis
...
+  c_text TEXT,
+  c_tinytext TINYTEXT,
+  c_mediumtext MEDIUMTEXT,
+  c_longtext LONGTEXT,
+  b_varbinary VARBINARY(10),
```
pali: Same as for VARCHAR

t/50chopblanks.t
urcheon: I've removed the VARCHAR and VARBINARY from the test case.
What I need to test is that binary fields are not being trimmed on the client side, and that's accomplished by BLOB and it's variants.

t/50chopblanks.t
urcheon: Test has been updated, and all automated tests pass

CaptTofu:
Hi there - could you please rebase? Travis builds were broken. I've merged a PR to fix it. I'd love to test this and get it merged.

urcheon:
Rebased, all tests pass

pali:
Patch was applied into https://github.com/gooddata/DBD-MariaDB repository.

mbeijen:
Thanks for your contribution!

PR #141, opened 2017-07-17
MYSQL_SECURE_AUTH became a no-op from MySQL 5.7.5 and is removed from MySQL 8.0.3.

berntm:
Needed to be able to compile with latest 8.0 source, but should be ignored from 5.7.5 since it is a no-op and has no effect there.


pali:
When running on libmysqlclient.so version which do not support this option, then it should die/croak and not silently ignore unsupported option.

CaptTofu:
seems to be failing AppVeyor. Could you please take a look and see what needs to be done to test this out? Thanks!

pali:
That bool change can cause problems for older mysql versions as API and ABI in those versions expects my_bool.

Why you are trying to invent wheel again and in wrong way? This problem with bools was already fixed in DBD::mysql 4.042 version.

berntm:
Can you pint me to the commit that fixes the bool problem for 4.042?

berntm:
Found it. The fix for my_bool (754e441c384269f6a4752059443341ca85e7720c) was reverted by 6fd72e55c9771b25b5775d08a120baae3181806e which basically reverts 4.042 back to 4.041 as new version 4.043.

pali:
https://github.com/perl5-dbi/DBD-mysql/pull/123

CaptTofu:
thank you!

PR #144, opened 2017-09-03
Add back improvements to CI testing

mbeijen:


pali:
GeomFromText was also fixed, but you reverted it.

PR #181, opened 2017-11-14
Upgrade Makefile.PL to meta-spec 2, github issues

Grinnz:
This updates the META_MERGE section in Makefile.PL to use the current meta-spec version 2, adds bugtracker metadata so github will be considered the canonical issue tracker, and sets dynamic_config to 0 since Makefile.PL does not set prereqs dynamically (this is only appropriate for meta-spec 2 - https://metacpan.org/pod/CPAN::Meta::Spec#dynamic_config).

I've verified that in the generated META.json this only adds more specific repository metadata, adds the bugtracker metadata, and sets dynamic_config to 0.

### DIFF
--- a/Makefile.PL
+++ b/Makefile.PL
@@ -376,12 +376,19 @@ if (eval $ExtUtils::MakeMaker::VERSION >= 5.43) {
     LICENSE => 'perl',
     MIN_PERL_VERSION => '5.008001',
     META_MERGE => {
+      'meta-spec' => { version => 2 },
+      dynamic_config => 0,
       resources => {
-        repository => 'https://github.com/perl5-dbi/DBD-mysql',
-        MailingList => 'mailto:dbi-dev@perl.org',
-        license     => 'http://dev.perl.org/licenses/',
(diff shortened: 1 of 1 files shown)


pali:
See: 816b100df0c8abcf27b2235283d672cb892dc358

@mbeijen @CaptTofu: it would be great if you process all reverted commits before merging in pull request, because after that it would be impossible.

Grinnz:
If the mentioned commit is re-applied, I can rebase this change to only add the bugtracker metadata, correct the license metadata to an array according to meta-spec 2, and set dynamic_config to 0.

pali:
@mbeijen @CaptTofu see all commits: https://github.com/perl5-dbi/DBD-mysql/compare/4.041...4.042

mbeijen:
Thanks @grinnz for your contribution!

@pali I fail to see how merging this would prevent other commits from being added again.

pali:
@mbeijen if you re-apply commits before merging new pull requests, then it would work...

Issue #120, opened 2017-04-13
DBD::mysql::st DESTROY failed on OpenBSD


glasswalk3r:
Some tests from DBD::mysql fail on OpenBSD with different perl versions (5.024001 and 5.020003). Here there are two different reports:

- http://www.cpantesters.org/cpan/report/53f3da90-17de-11e7-a4a0-8eacd02b4708
- http://www.cpantesters.org/cpan/report/26042072-17de-11e7-a4a0-8eacd02b4708

Both failed with the same errors:

```
DBD::mysql::st DESTROY failed: ßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßßß
```

And this "beta" character goes on in very long lines.

It is important to note that MySQL is not indeed available on **OpenBSD 6.0** by default, but MariaDB is installed instead:

```
# mysql_clientinfo is: 10.0.25-MariaDB
# mysql_clientversion is: 100025
# mysql_serverversion is: 100025
# mysql_hostinfo is: Localhost via UNIX socket
# mysql_serverinfo is: 10.0.25-MariaDB
# mysql_stat is: Uptime: 1806  Threads: 1  Questions: 7763  Slow queries: 0  Opens: 430  Flush tables: 1  Open tables: 62  Queries per second avg: 4.298
# mysql_protoinfo is: 10
# SQL_DBMS_VER is 10.0.25-MariaDB
# Default storage engine is: InnoDB
```

I also have reports from this same CPAN Smoker with PASS and UNKNOWN results, but in both cases the tests were skipped.

pali:
Hi! Look at output from test `t/40server_prepare_crash.t`:
```
# Your libmysqlclient.so is vulnerable to CVE 2017-3302 and can crash perl
```

[...]

PR #222, opened 2017-12-28
Fixed tests that cannot be executed on OpenBSD

glasswalk3r:
Due dependency on Proc::ProcessTable, the tests changed in this commit
cannot be executed successfully on OpenBSD, thus they will be disabled to
run on that platform until a workaround is available.

Tests were changed based on patch provided by @pali .

For more details, see https://github.com/perl5-dbi/DBD-mysql/issues/120


mbeijen:
Thanks for your contribution!

PR #130, opened 2017-05-23
Run Travis tests on Perl 5.8.1 builds from unofficial travis-perl project

pali:

PR #225, opened 2018-01-14
Ssl reapply fix

CaptTofu:


dbdimp.c
```diff
@@ -1903,28 +1909,34 @@ MYSQL *mysql_dr_connect(
...
+	    if ((svp = hv_fetch(hv, "mysql_ssl_verify_server_cert", 28, FALSE)) && *svp) {
+  #if defined(HAVE_SSL_VERIFY) || defined(HAVE_SSL_MODE)
+	      ssl_verify = SvTRUE(*svp);
+	      ssl_verify_set = 1;
+  #else
+	      set_ssl_error(sock, "mysql_ssl_verify_server_cert=1 is not supported");
+	      return NULL;
+  #endif
+	    }
+        if ((svp = hv_fetch(hv, "mysql_ssl_optional", 18, FALSE)) && *svp)
```
dveeden: This throws a compiler warning with GCC because if misleading indentation 

```
dbdimp.c:1937:9: warning: this ‘if’ clause does not guard... [-Wmisleading-indentation]
         if ((svp = hv_fetch(hv, "mysql_ssl_optional", 18, FALSE)) && *svp)
```

dbdimp.c
```diff
@@ -1903,28 +1909,34 @@ MYSQL *mysql_dr_connect(
...
-              the certificate
-            */
-	    my_bool ssl_verify_true = 0;
-	    if ((svp = hv_fetch(hv, "mysql_ssl_verify_server_cert", 28, FALSE))  &&  *svp)
-	      ssl_verify_true = SvTRUE(*svp);
-#endif
+	    unsigned int ssl_mode;
+	    my_bool ssl_enforce = 1;
+	    my_bool ssl_verify = 0;
+	    my_bool ssl_verify_set = 0;
```
dveeden: This also gives a compiler warning from GCC
```
dbdimp.c:1925:14: warning: variable ‘ssl_verify_set’ set but not used [-Wunused-but-set-variable]
      my_bool ssl_verify_set = 0;
```

I think a `#ifdef defined(HAVE_SSL_VERIFY)` should be placed around this. But otherwise it is harmless.

dveeden:
When building with `--nossl` the tests fail:
```
t/92ssl_backronym_vulnerability.t ....... ok   
t/92ssl_optional.t ...................... DBI connect('test:127.0.0.1:5718','msandbox',...) failed: SSL connection error: mysql_ssl=1 is not supported at t/92ssl_optional.t line 18.
t/92ssl_optional.t ...................... 1/2 
#   Failed test 'DBD::mysql supports mysql_ssl_optional=1 and connect via plain text protocol when SSL is not supported by server'
#   at t/92ssl_optional.t line 19.
# Error code: 2026
# Error message: SSL connection error: mysql_ssl=1 is not supported
DBI connect('test:127.0.0.1:5718','msandbox',...) failed: SSL connection error: mysql_ssl=1 is not supported at t/92ssl_optional.t line 21.

#   Failed test 'DBD::mysql supports mysql_ssl_optional=1 and connect via plain text protocol when SSL is not supported by server even with mysql_ssl_ca_file'
#   at t/92ssl_optional.t line 22.
# Error code: 2026
# Error message: SSL connection error: mysql_ssl=1 is not supported
# Looks like you failed 2 tests of 2.
t/92ssl_optional.t ...................... Dubious, test returned 2 (wstat 512, 0x200)
Failed 2/2 subtests 
t/92ssl_riddle_vulnerability.t .......... ok   
```
Probably best to make sure that they are skipped.

pali:
> Probably best to make sure that they are skipped.

Test looks like to be correct. When `mysql_ssl_optional=1` is set then SSL/TLS is not required and only optional. So problem is in driver.

dbdimp.c
CaptTofu: thanks - I'll look at that.

CaptTofu: gah, there si some 4 space indentation, some 2. Not sure what we all agreed upon. MySQL source when I worked there was always 2. Thoughts?

CaptTofu: also, this doesn't look this way in my editor. I need to figure out what .vimrc to get me what I want.

dveeden:
An initial fix clients without SSL failing connections with optional SSL:
https://github.com/dveeden/DBD-mysql/commit/e200a47e1fbe14b39dfbdfede318440b826225cf

dveeden:
Some of the travis failures is perlbrew not succeeding to install a specific version.

dveeden:
This commit seems to fix the build with Perl 5.8:
https://github.com/dveeden/DBD-mysql/commit/76bb2e6aea76f68901986f809a816116be530c4b

pali:
> This commit seems to fix the build with Perl 5.8 ...

What about looking into git history before trying to reinvent wheel? :-) It was already fixed:
https://github.com/perl5-dbi/DBD-mysql/commit/0dad3d251a3f466f511815ce88f10f2f3143274b
https://github.com/perl5-dbi/DBD-mysql/pull/130

dveeden:
You're right.
Any other commits that should be re-applied to get TLS support fixed?

PR #226, opened 2018-01-14
Allow testsuite to run against 8.0.3

dveeden:
Two issues:

1. `AsBinary` and `GeomFromText` are gone in 8.0. Use `ST_AsBinary` and `ST_GeomFromTest` instead.
2. MySQL 8.0.3 returns too many rows for `DESCRIBE` with a column specification

https://bugs.mysql.com/bug.php?id=89224

### DIFF
--- a/t/55utf8.t
+++ b/t/55utf8.t
@@ -55,16 +55,27 @@ cmp_ok $dbh->quote($blob), 'eq', $quoted_blob, 'testing quoting of blob';
 #ok $dbh->{mysql_enable_utf8}, "mysql_enable_utf8 survive connect()";
 $dbh->{mysql_enable_utf8}=1;
 
-my $query = <<EOI;
+my $query;
+if ($dbh->{mysql_clientversion} < 50600) {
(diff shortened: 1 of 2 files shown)


pali:
It was already fixed in 41de8940b6c623d87bf8f0fb679317e07056447b and 0626173557514519cc0520dde19cf67e3b84d212.

dveeden:
You're right. 
7ba0a0331e3682b33870f347ba7794a7acfaaa8c seems to fix the 2nd issue.
However fixes for both issues were reverted.

mbeijen:
The other fix is better, the availability of `GeomFromText` does NOT depend on the client version, but on the server version instead!

Also, depending the version number feels so much as 'user agent sniffing' to me; and I'd like to avoid that if possible. Especially with the MariaDB version number as well. I'm aware we do that all the time, though.

Is there not a nice idiomatic way to check the availability of the function without depending of the version number? I also asked here: https://dba.stackexchange.com/questions/197751/mysql-test-if-function-is-available

PR #245, opened 2018-03-08
Update docs: Multithreading support

### LINKED ISSUE #244: Docs: Statement about multithreading is outdated
From http://search.cpan.org/~capttofu/DBD-mysql-4.046/lib/DBD/mysql.pm

```
MULTITHREADING ^

The multithreading capabilities of DBD::mysql depend completely on the underlying C libraries. The modules are working with handle data only, no global variables are accessed or (to the best of my knowledge) thread unsafe functions are called. Thus DBD::mysql is believed to be completely thread safe, if the C libraries are thread safe and you don't share handles among threads.

The obvious question is: Are the C libraries thread safe? In the case of MySQL the answer is "mostly" and, in theory, you should be able to get a "yes", if the C library is compiled for being thread safe (By default it isn't.) by passing the option -with-thread-safe-client to configure. See the section on How to make a threadsafe client in the manual.
```

Isn't libmysqlclient always thread safe now?

https://mysqlserverteam.com/deprecating-the-mysql_config-libs_r-option/ suggests it is thread safe by default since 5.5

dveeden:
Fixes #244

### DIFF
--- a/lib/DBD/mysql.pm
+++ b/lib/DBD/mysql.pm
@@ -1473,7 +1473,7 @@ against:
 
 =item mysql_dbd_stats
 
-  $info_hashref = $dhb->{mysql_dbd_stats};
+  $info_hashref = $dbh->{mysql_dbd_stats};
 
 DBD::mysql keeps track of some statistics in the mysql_dbd_stats attribute.
 The following stats are being maintained:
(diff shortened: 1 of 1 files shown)


mbeijen:
Thanks! 👍

PR #247, opened 2018-03-14
Fix a SEGV when using mariadb connector

atoomic:
When using mariadb connector 3.0.3 with DBD::mysql
we can notice a segfault on reconnection due
to the fact that the mysql extensions are not reset.

Doing the extra init as performed by DBD::mariadb
with some extra save and restore fixes the issue.

### DIFF
--- a/dbdimp.c
+++ b/dbdimp.c
@@ -2371,23 +2371,48 @@ int dbd_db_disconnect(SV* dbh, imp_dbh_t* imp_dbh)
 #ifdef dTHR
   dTHR;
 #endif
+
+  /* patched from mariadb_db_disconnect to clear mysql.extension correctly */
+
   dTHX;
   D_imp_xxh(dbh);
+  const void *methods;
+  char last_error[sizeof(imp_dbh->pmysql->net.last_error)];
+  char sqlstate[sizeof(imp_dbh->pmysql->net.sqlstate)];
+  unsigned int last_errno;
 
   /* We assume that disconnect will always work       */
   /* since most errors imply already disconnected.    */
   DBIc_ACTIVE_off(imp_dbh);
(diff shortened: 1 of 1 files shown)


atoomic:
note: that appveyor CI issue seems not related to the issue
```
Building Time-HiRes-1.9755 ... ! Installing Time::HiRes failed. See C:\Users\appveyor\.cpanm\work\1521052705.1676\build.log for details. Retry with --force to force install it.
166FAIL
```

atoomic:
error discussed upstream here:  https://jira.mariadb.org/browse/CONC-316

pali:
This patch is of course incorrect/incomplete and not for production usage. I created it for DBD::MariaDB just as quick fix until I fix it properly.

atoomic:
@pali thanks for your input, I understand that this code is not optimized and might be incomplete, when trying to restore the context on a freshly created connection, but I think in most cases disconnect/reconnect should not be frequent, and clearly prefer a full reconnection rather than a segfault at this point.

What other issues do you have in mind? Any known problems with this approach outside performance?

Would you not consider using mariadb connector 3.0.3 with DBD::mysql? Would you prefer using DBD::MariaDB instead?

thanks

pali:
I will try to fix this problem properly without touching internal MYSQL structures... basically it is needed for proper support when compiled against new version of MariaDB Connector/C.

For production code I suggest to stick with old client libraries MySQL 5.6/5.7 or MariaDB 10.0, both are still supported by upstream.

For testing you can use DBD::MariaDB with Connector/C 3.x or MariaDB 10.3, but be aware that there can be changes in future, prior uploading version to CPAN.

dveeden:
Closing this as this is now handled in #291 ( dc32f84ce14272ea760c805e84c4bd259edf5e25 ).

PR #276, opened 2018-10-07
Fix Travis Config

### LINKED ISSUE #268: Travis-CI setup broken
Related: https://github.com/perl5-dbi/DBD-mysql/issues/262#issuecomment-421616497

Looks like 6fd72e55c9771b25b5775d08a120baae3181806e ("New version 4.043 - the same as 4.041") removed some critical bits of the `.travis-ci.yaml` file causing tests to always run against the system version of MySQL instead of the specified MySQL or MariaDB version.
These were originally introduced in 26fbef8a67bd8c3954d376dfb424261943752209

We should fix the `before_install` etc.

dveeden:
Builds seem to be broken for:
* Perl 5.18-shrplib
* MySQL 4.1.22
* MySQL 5.0.15
* MySQL 8.0.12 (not a build issue,  but an issue with setting up the sandbox)
* MariaDB 10.3.9

Fixes #268

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,28 @@
+sudo: false
+
+cache:
+  directories:
+    - $HOME/cache
+    - $HOME/.cpan/sources
+
 language: perl
+
 perl:
+  - "5.8.1"
   - "5.8"
   - "5.10"
+  - "5.12.0"
   - "5.12"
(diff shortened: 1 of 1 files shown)


CaptTofu:
Merging to fix build

PR #284, opened 2018-11-04
Move Travis to dbdeployer

dveeden:
This is because MySQL::Sandbox is deprecated.


dveeden:
This doesn't work for 4.1 and old 5.0 releases yet.

But there is a fix: https://github.com/datacharmer/dbdeployer/pull/40

PR #296, opened 2019-01-10
Make type of is_null field consistent and same as in MYSQL_BIND

ericherman:
This is a `git cherry-pick` 5cd65be3cca912dba90507f8bf10fd966b8a79ed because this seems to have gotten lost as part of 6fd72e55c9771b25b5775d08a120baae3181806e

Fixes compile warning: dbdimp.c:3340:24: warning: assignment from incompatible pointer type [enabled by default]


ericherman:
See Also:

https://github.com/mysql/mysql-server/blob/8.0/include/mysql.h#L621
https://github.com/MariaDB/server/blob/10.4/include/mysql.h#L694
https://github.com/perl5-dbi/DBD-mysql/blob/master/dbdimp.h#L66 ( introduced in 850fc49b )

PR #303, opened 2019-02-25
Mention default and reword mysql_client_found_rows POD for clarity

wolfsage:


wolfsage:
Really I'd just like a default value mentioned, but I think this more clear than what was there before.

dveeden:
Thank you!

PR #308, opened 2019-03-06
Fix for the zerofil issue on MariaDB 10.3.13

### LINKED ISSUE #304: libdbd-mysql-perl: FTBFS against mariadb-10.3 1:10.3.13-1
We have the following bug reported to the Debian package of
DBD-mysql, c.f. https://bugs.debian.org/923541

It doesn't seem to be a bug in the packaging, so you may want to take
a look. Thanks!

AFAICS, DBD-mysql is tested against all kinds of MySQL and MariaDB versions
on travis but apparently not against 10.3.13.  Maybe this explains why this
issue hasn't popped up there yet.

------8<-----------8<-----------8<-----------8<-----------8<-----

```
Source: libdbd-mysql-perl
Version: 4.050-1
Severity: serious
Tags: upstream ftbfs sid
Justification: fails to build from source (but built successfully in the past)

This is already discussed in
https://lists.debian.org/debian-perl/2019/02/msg00026.html
but filing it as a bug:

libdbd-mysql-perl fails its tests with

t/rt118977-zerofill.t ................... 
1..8
ok 1
ok 2
ok 3
ok 4
ok 5
ok 6
ok 7
not ok 8
Dubious, test returned 1 (wstat 256, 0x100)
Failed 1/8 subtests 

when built against mariadb-10.3 1:10.3.13-1.

This looks like #917303 which was fixed in 4.050-1 and this fix
worked with mariadb-10.3 1:10.3.12-2. Seems that there is a change in
mariadb between 10.3.12 and 10.3.13 which breaks the test again.


Cheers,
gregor


```
------8<-----------8<-----------8<-----------8<-----------8<-----

Thanks for considering,
  gregor herrmann,
  Debian Perl Group

dveeden:
Closes #304

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -128,6 +128,8 @@ matrix:
     - perl: "5.20"
       env: DB=MariaDB VERSION=10.3.11
     - perl: "5.20"
+      env: DB=MariaDB VERSION=10.3.13
+    - perl: "5.20"
       env: DB=MariaDB VERSION=10.4.1
     - perl: "5.20"
       env: CONC_DB=MySQL CONC_VERSION=6.0.0-beta
--- a/.travis.yml
+++ b/.travis.yml
@@ -185,8 +185,11 @@ before_install:
       esac ;
       SANDBOX_CACHE_FILE="$HOME/cache/$(basename "$SANDBOX_URL")" ;
     elif [ "$DB" = "MariaDB" ]; then
-      SANDBOX_URL=https://downloads.mariadb.com/MariaDB/mariadb-$VERSION/bintar-linux-x86_64/mariadb-$VERSION-linux-x86_64.tar.gz ;
-      SANDBOX_CACHE_FILE="$HOME/cache/mariadb-$VERSION-linux-x86_64.tar.gz" ;
+      case "$VERSION" in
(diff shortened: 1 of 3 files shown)


gregoa:
8b5ba5f9c8d239328ecbc862aba203d44819e2f5 indeed works on Debian with MariaDB 10.3.13 and also MySQL 5.7.24. Great!

If I add 492e2a62c4f34f967a099d736a264e2f904f5f30 on top of it, it fails with MariaDB 10.3.13:
```
#   Failed test at t/rt118977-zerofill.t line 22.
#          got: '1'
#     expected: '00001'
# Looks like you failed 1 test of 8.
t/rt118977-zerofill.t ................... 
1..8
ok 1
ok 2
ok 3
ok 4
ok 5
ok 6
ok 7
not ok 8
Dubious, test returned 1 (wstat 256, 0x100)
Failed 1/8 subtests 
```
(and the same for MySQL 5.7.24).
Looks like the MariaDB failure is also on Travis, and MySQL 5.7.24 isn't tested there.

dveeden:
In master compiling against MariaDB client libs is no longer supported. Use DBD::MariaDB for MariaDB.

PR #313, opened 2019-06-13
Fix type conversions

### LINKED ISSUE #78: Problems with type conversions
Hi,
after upgrading to DBD::mysql 4.041 a problem appeared in amavisd-new:
https://lists.amavis.org/pipermail/amavis-users/2016-December/004674.html
The problem could be related to this commit:
https://github.com/perl5-dbi/DBD-mysql/commit/caea0b774028650c0cbd9d8f9c4a0b47831116df
 Cheers
  Giovanni

### LINKED ISSUE #312: BIGINT type error conversion with amavisd-new
#78 is still not fixed.

It still does not work with DBD-MySQL 4.050, Perl 5.16.3 and amavisd-new-2.11.1 (20181009) - foreign key constraint fails because of BIGINT type-error.

Rolling back to 4.037 does work, so it is due to a change in DBD-MySQL

tomsommer:
Calling SvNV() for magical scalar is not enough for float type conversion.
It caused problem for Amavis in tainted mode -- all float values were zero.
On the other hand SvIV() and SvUV() seems to work fine. To be sure that
correct value of float is in scalar use sv_setnv() with explicit NV float
value. Similar code is changed also for integers IV/UV.

Fixes #78 
Fixes #312 

Credit https://github.com/kentnl-gentoo/DBD-mysql/commit/b6b8540216bd03b68f3bc076b3d3106f4be23f9d

### DIFF
--- a/dbdimp.c
+++ b/dbdimp.c
@@ -4447,8 +4447,7 @@ process:
           if (!(fields[i].flags & ZEROFILL_FLAG))
           {
             /* Coerce to double and set scalar as NV */
-            (void) SvNV(sv);
-            SvNOK_only(sv);
+            sv_setnv(sv, SvNV(sv));
           }
           break;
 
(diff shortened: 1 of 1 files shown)


tomsommer:
Partially reverts commit from https://github.com/perl5-dbi/DBD-mysql/commit/0200483708e46d16293868dd974b91ec51637e11 although https://github.com/perl5-dbi/DBD-mysql/commit/caea0b774028650c0cbd9d8f9c4a0b47831116df is known for blame

PR #324, opened 2020-12-10
Added some basic docker support for building and testing

rykr:


dveeden:
While this could be considered as new functionality it has basically zero chance of breaking existing stuff, so merging this.

rykr:
Thanks!

Issue #329, opened 2021-05-12
rt85919-fetch-lost-connection.t failing on MySQL 8.0.24+ due to new error code


sharris40:
See https://dev.mysql.com/doc/mysql-errors/8.0/en/server-error-reference.html#error_er_client_interaction_timeout

The test expects error code 2006 or 2013, both of which are generated by the client library. Since 8.0.24, we receive the error code 4031, `ER_CLIENT_INTERACTION_TIMEOUT`, which is generated by the server. This new code causes the test to fail.

PR #330, opened 2021-06-10
accept 4031 as a valid return code when connection closed

waltman:
As of 8.0.24 MySQL writes the reason the connection was closed before
closing it, so 4031 (ER_CLIENT_INTERACTION_TIMEOUT) is now an valid
return code. See

https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-24.html#mysqld-8-0-24-connection-management

for more information.

This addresses the issue reported in https://github.com/perl5-dbi/DBD-mysql/issues/329


dveeden (approved):

robbiebow:
Hi, is there an ETA on releasing this fix?

PR #349, opened 2022-11-23
fix some queries with columns beginning with `limit` would fail

kebhr:
#288 fixes a problem in which SQL involving columns whose names end in `limit` cannot be generated successfully.
However, this fix caused SQL containing columns starting with `limit` to fail to generate successfully (e.g. `SELECT * FROM table WHERE limit_by > ?`).
This is because the syntax interpretation compares it to `limit` when it should be compared to `limit ` which includes white space.

I have created a fix for this problem and a test code that can confirm this issue.
With 4.050, modified [t/35limit.t](https://github.com/perl5-dbi/DBD-mysql/blob/4f6e91178db41cce4920680800d7a72d9ebe11ff/t/35limit.t) fails, but when this pull request is applied, it succeeds.

### DIFF
--- a/dbdimp.c
+++ b/dbdimp.c
@@ -605,8 +605,8 @@ static char *parse_params(
         it would be good to be able to handle any number of cases and orders
       */
       if (((*statement_ptr == ' ') || (*statement_ptr == '\n') || (*statement_ptr == '\t')) &&
-          (!strncmp(statement_ptr+1, "limit ", 5) ||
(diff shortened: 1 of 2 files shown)


kebhr:
@dveeden Hello. Could you please take a look at this pull-request when you have a moment?

mohawk2:
This seems ripe for merging - it has tests for the additional behaviour?

dveeden (approved):

Issue #354, opened 2023-07-18
As of MySQL 8.0.34 and 8.0.10 's libmysqlclients deprecates `MYSQL_OPT_RECONNECT`


yoku0825:
```
The MySQL client library currently supports performing an automatic reconnection to the server if it finds that the connection is down and an application attempts to send a statement to the server to be executed. Now, this feature is deprecated and subject to removal in a future release of MySQL.

The related MYSQL_OPT_RECONNECT option is still available but it is also deprecated. C API functions [mysql_get_option()](https://dev.mysql.com/doc/c-api/8.1/en/mysql-get-option.html) and [mysql_options()](https://dev.mysql.com/doc/c-api/8.1/en/mysql-options.html) now write a deprecation warning to the standard error output when an application specifies MYSQL_OPT_RECONNECT. (WL #15766)
```

https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-34.html#mysqld-8-0-34-deprecation-removal
https://dev.mysql.com/doc/relnotes/mysql/8.1/en/news-8-1-0.html#mysqld-8-1-0-deprecation-removal

I seem DBD::mysql is using only one place in

https://github.com/perl5-dbi/DBD-mysql/blob/7117e4c53fdd4d1ddd6672e339723bd5b3bd73ec/dbdimp.c#L2118-L2121

Current status is `deprecated` and it would work several minor updates.
But the after the deadline, this method raises would be errored.
(This depends libmysqlclient.so 's version, not Server's version)

philip:
It's worth noting that setting it to false also emits an error, so DBD::mysql always causes this error as of 8.0.34/8.1.0:

    WARNING: MYSQL_OPT_RECONNECT is deprecated and will be removed in a future version.

maxbarry:
Ubuntu rolled out `mysql 8.0.34-0ubuntu0.22.04.1` overnight and I'm getting flooded with these errors from cron-triggered Perl scripts.

youradds:
Same issue as of last night! Having to set the crons to devnull, otherwise I'm getting spammed every 1 minute the cron runs :(

daviding58:
Same here as of today with Ubuntu 20.04.
In my opinion that warning should be given by the MySQL code **only** if a call tries to set the option to TRUE.
MySQL's **own** website recommends setting it to FALSE [here](https://dev.mysql.com/doc/c-api/8.0/en/c-api-auto-reconnect.html)

[...]

PR #355, opened 2023-08-16
MYSQL_OPT_RECONNECT is deprecated in MySQL 8.0.34.  Removed support.

aorr436:
Removed support for MYSQL_OPT_RECONNECT for MySQL versions >= 8.0.34

https://github.com/perl5-dbi/DBD-mysql/issues/354

### DIFF
--- a/dbdimp.c
+++ b/dbdimp.c
@@ -2115,7 +2115,7 @@ MYSQL *mysql_dr_connect(
         we turn off Mysql's auto reconnect and handle re-connecting ourselves
         so that we can keep track of when this happens.
       */
-#if MYSQL_VERSION_ID >= 50013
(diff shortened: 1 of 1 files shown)


minusdavid:
Noting that Ubuntu have added this patch to their official packages: https://bugs.launchpad.net/ubuntu/+source/mysql-8.0/+bug/2031548

tonvoon:
We had to patch this as

`#if defined(MARIADB_BASE_VERSION) || MYSQL_VERSION_ID >= 50013 && MYSQL_VERSION_ID < 80034`

otherwise it failed to compile properly on MariaDB 5.5.

dveeden:
This is fixed in master now

PR #422, opened 2024-03-20
docs: Update package names and MariaDB related things

dveeden:
- Fedora now has both MySQL and MariaDB packages (more changes might be coming in the future: https://lwn.net/Articles/960630/ )
- DBD::mysql requires MySQL 8.x
- MariaDB 5.2.0 is very old (a lot has happened since)

@michal-josef-spacek I found this while checking #420 , maybe you can give this a quick check?

### DIFF
--- a/lib/DBD/mysql.pm
+++ b/lib/DBD/mysql.pm
@@ -1376,19 +1376,19 @@ against:
 
   print "$dbh->{mysql_clientinfo}\n";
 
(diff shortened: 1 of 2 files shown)


michal-josef-spacek:
@dveeden I tested on Fedora 39
* Installation is ok
* MariaDB [test] → MySQL [test]

PR #432, opened 2024-06-13
Fix calls to mysql_server_end()/mysql_library_end()

### LINKED ISSUE #431: SSL problems after locally scoped disconnect
> See the note at the end of this report first...

### DBD::mysql version

5.003, 5.004, 5.005

### MySQL client version

8.0.37

### Server version

8.4.0

### Operating system version

Linux (amazonlinux:2), libssl.so.1.0.2k

### What happened?

Creating a connection using `DBD::mysql` versions 5.003 - 5.005 in a locally scoped block and **_then disconnecting in the same locally scope block_** puts SSL in a state where downstream SSL connections fail with a 500 (internal response).

The script below will demonstrate the problem:

```
use strict;
use warnings;

use LWP::UserAgent;
use Net::SSLeay;
use HTTP::Request;
use Data::Dumper;
use DBI;

$Net::SSLeay::trace = 2;

my $dsn = sprintf 'dbi:mysql:%s:%s', $ENV{DBI_DBNAME}, $ENV{DBI_HOST};

my $dbi = DBI->connect( $dsn, $ENV{DBI_USER}, $ENV{DBI_PASS} );

my $req = HTTP::Request->new( GET => 'https://google.com' );

my $ua = LWP::UserAgent->new;

my $rsp = $ua->request($req);

print {*STDERR} Dumper( [ status => $rsp->status_line ] );

connect_and_do_something($dsn);

$rsp = $ua->request($req);

print {*STDERR} Dumper( [ status => $rsp->status_line ] );

########################################################################
sub connect_and_do_something {
########################################################################
    my ($dsn) = shift;

    my $dbi = DBI->connect( $dsn, $ENV{DBI_USER}, $ENV{DBI_PASS} );
    $dbi->disconnect;
}

1;
```

The problem persists regardless of the 8.x MySQL client library I use to build `DBD::mysql` (I've tried several), so I believe the problem is related to what has been changed from version 5.002 to version 5.003. The problem does not manifest itself in  5.002 or when using version 4.050.

The output of the program above is attached:

[bug-report.txt](https://github.com/user-attachments/files/15821982/bug-report.txt)

### Other information

In version 5.003 `mysql_library_end()` was added.  Removing this line "fixes" the issue, although there may be other side effects of removing this line (memory leaks?).  At the very least this sub is probably being called prematurely as it appears to remove plugins (like openssl plugin that apparently resets the TLS context):

```
static void deinit(mysql_harness::PluginFuncEnv *) {
  // let the TlsLibraryContext destructor do the SSL cleanup
  tls_library_context.reset();
}
```

Incidentally, it appears that `dbd_db_destroy()` is being invoked when a database handle goes out scope in either a closure or a sub. This implies you probably should not be calling `mysql_library_end()` in that subroutine?

`dbd_discon_all` might be the appropriate place to call `mysql_library_end()`  but it appears to have been disabled back in 2003.  You should also know that `mysql_server_end()`  is called in ` dbd_discon_all()` already.  (`mysql_server_end` is an alias for `mysql_library_end()`).

dveeden:
- Stop calling `mysql_library_end()` in `dbd_db_destroy()` as this leads to issues with OpenSSL in specific circumstances.
- Call `myql_library_end()` instead of `mysql_server_end()` as the latter has been deprecated.

See also:
- https://dev.mysql.com/doc/c-api/8.4/en/mysql-server-end.html
- https://dev.mysql.com/doc/c-api/8.4/en/mysql-library-end.html

Closes #431

### DIFF
--- a/dbdimp.c
+++ b/dbdimp.c
@@ -1827,7 +1827,7 @@ int dbd_discon_all (SV *drh, imp_drh_t *imp_drh) {
   dTHX;
   PERL_UNUSED_ARG(drh);
 
-  mysql_server_end();
+  mysql_library_end();
 
   /* The disconnect_all concept is flawed and needs more work */
   if (!PL_dirty && !SvTRUE(perl_get_sv("DBI::PERL_ENDING",0))) {
@@ -1871,7 +1871,6 @@ void dbd_db_destroy(SV* dbh, imp_dbh_t* imp_dbh) {
     }
     dbd_db_disconnect(dbh, imp_dbh);
   }
-  mysql_library_end();
   Safefree(imp_dbh->pmysql);
 
   /* Tell DBI, that dbh->destroy must no longer be called */
--- a/dbdimp.h
+++ b/dbdimp.h
@@ -237,6 +237,7 @@ struct imp_sth_st {
 #define dbd_db_destroy		mysql_db_destroy
 #define dbd_db_STORE_attrib	mysql_db_STORE_attrib
 #define dbd_db_FETCH_attrib	mysql_db_FETCH_attrib
+#define dbd_discon_all		mysql_discon_all
 #define dbd_st_prepare		mysql_st_prepare
 #define dbd_st_execute		mysql_st_execute
 #define dbd_st_fetch		mysql_st_fetch


dbdimp.c
```diff
@@ -1827,7 +1827,7 @@ int dbd_discon_all (SV *drh, imp_drh_t *imp_drh) {
   dTHX;
```
rlauer6, line 1827: Unfortunately I don't think this subroutine is actually called any more.  It was removed in 2003 as per commit history?  I think that function needs to be re-visited and must be called to call the `mysql_library_end()` sub.  This is actually quite a bad situation as `DBD::mysql` is currently not cleaning up at the right point, at least in my reading of the situation.  I am fairly new to looking at this, so I defer to those with more insight, however empirically I know that calling `mysql_library_end` in every call to DESTROY is causing issues when the `openssl_plugin` is unloaded.

dveeden: I've added this in `dbdimp.h` and it seems to get called. So this is one step forward.

Looks like 3921e6bee0e5480c9490cdff66c4c190c085bb44 was the commit that removed this.

rlauer6: Just curious how you verified that it is being called?  I did the same thing by adding it to `dbdimp.h` and then adding some print statements (I know that is pretty crude) but did not see any output.  Adding print statements at other points in the C code produces messages on my console, so I'm pretty sure that if it had been called I would have seen the messages.  OTOH - I may have made some other error that prevented me from seeing the message.  In any case, I do think this  is where `mysql_library_end` needs to be called but would like to know how we can verify that.  Thanks so much...nice work.

dveeden: Adding something like `croak("this is called")`  seems to work for me and so does GDB with adding a breakpoint ( `b mysql_discon_all` iirc).

rlauer6: perfect - thank you

PR #384, opened 2023-11-01
Try new appveyor.yml

dveeden:
Closes #359

Issue #359, opened 2023-08-25
Setup CI job for Windows and MySQL 8.x


dveeden:
GitHub has windows runners, but it looks like MySQL requires a quite old Visual C++ version according to https://dev.mysql.com/doc/c-api/8.0/en/c-api-building-clients.html

dveeden:
I tried this, but that results in msvcrt issues...
```
  test-windows:
    runs-on: windows-2019
    steps:
      - uses: actions/checkout@v3
      - name: Set up perl
        uses: shogo82148/actions-setup-perl@v1
        with:
          perl-version: 5.36
          install-modules-with: cpanm
          install-modules-args: --with-develop --with-suggests
          install-modules: Devel::CheckLib DBI::DBD
      - run: choco install mysql
      - name: Build
        run: |
          perl Makefile.PL --testhost=127.0.0.1 --testuser=root --cflags='-I"C:\tools\mysql\current\include"' --libs='-L"C:\tools\mysql\current\lib" -lmysqlclient'
          gmake
      - name: Test
        run: gmake test
```

dveeden:
In the v4 branch there is `appveyor.yml`, but looks like appveyor doesn't support MySQL 8.0 ( see https://www.appveyor.com/docs/services-databases/#mysql )

```yml
version: 1.0.{build}

services:
  - mysql

install:
  - cinst StrawberryPerl
  - path C:\strawberry\perl\bin;C:\strawberry\perl\site\bin;C:\strawberry\c\bin;%PATH%
  - mkdir %APPVEYOR_BUILD_FOLDER%\tmp
  - set TMPDIR=%APPVEYOR_BUILD_FOLDER%\tmp
  - perl -V
  - cpan App::cpanminus
  - cpanm -q --showdeps --with-develop --with-suggests . | findstr /v "^perl\>" | cpanm -n
  - cpanm -q -n Devel::CheckLib

build_script:
  - perl Makefile.PL --mysql_config=c:\strawberry\c\bin\mysql_config.bat --testuser=root --testpassword=Password12!

test_script:
  - gmake test
```

dveeden:
Related: https://github.com/appveyor/ci/issues/3834

dveeden:
@Tux do you have the right permissions to disable/remove appveyor? Getting DBD::mysql to work on Windows is unfortunately not going to be solved any time soon I'm afraid.

PR #444, opened 2024-09-18
Make `statistics_info` use catalog arg

mohawk2:
Sorry for taking two goes at this. It was bothering me today that I saw the catalog info coming back from the db, but I knew `statistics_info` had filtering on that disabled. Now it's not (with a test, and everything!).

### DIFF
--- a/lib/DBD/mysql.pm
+++ b/lib/DBD/mysql.pm
@@ -738,12 +738,10 @@ EOF
     my @where;
     my @bind;
 
-    # catalogs are not yet supported by MySQL
-
-#    if (defined $catalog) {
-#        push @where, 'TABLE_CATALOG = ?';
-#        push @bind, $catalog;
-#    }
+    if (defined $catalog) {
+        push @where, 'TABLE_CATALOG = ?';
+        push @bind, $catalog;
+    }
 
     if (defined $schema) {
         push @where, 'TABLE_SCHEMA = ?';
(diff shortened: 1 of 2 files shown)


dveeden:
As MySQL doesn't support catalogs, the name it returns is always `def` isn't it? Does it make sense to do filtering on that?

```
mysql-9.0.1> SELECT CATALOG_NAME FROM information_schema.schemata GROUP BY CATALOG_NAME;
+--------------+
| CATALOG_NAME |
+--------------+
| def          |
+--------------+
1 row in set (0.01 sec)
```

I know that there are plans to add catalog support to MariaDB, but that hasn't been merged yet.

- https://mariadb.com/kb/en/catalogs-overview/

mohawk2:
It seems to me that in that case this addition is harmless, and if someone uses this DBD for MariaDB (which I've heard of as a thing) after it gets catalogs, this addition would become valuable?

mohawk2:
By the way, your AppVeyor config looks broken - it's giving an error talking about not finding project files?

dveeden:
> By the way, your AppVeyor config looks broken - it's giving an error talking about not finding project files?

Yes, this is a known issue. See also:
- https://github.com/perl5-dbi/DBD-mysql/issues/359
- https://github.com/perl5-dbi/DBD-mysql/pull/384

Basically AppVeyor didn't have MySQL 8.x (it now does) but the Strawberry Perl is still build with MySQL 5.7.x client libraries, which are too old for DBD::mysql. I would still like to get this working again to be able to test on windows.

PR #445, opened 2024-09-19
Test for LIMIT case insensitive

dveeden:
The test was testing for `LIMIT` and `limit` but didn't work for `Limit` etc.

Follow up for https://github.com/perl5-dbi/DBD-mysql/pull/349

@kebhr @mohawk2 what do you think about this?

I think `strncasecmp` may not be available on Windows, so this might need an `ifdef` with a Windows specific solution.


mohawk2:
I think you're right that it should be case-insensitive! And my quick googling shows there is indeed a `#ifdef` needed on Windows.

kebhr:
I also agree to this change. It will follow the MySQL specification more closely and do not believe that this change will cause a degrade.

PR #448, opened 2024-10-24
Tidy build error on Debian bookworm

davel:
We fail to build on Debian bookworm, it supplies MariaDB rather than MySQL. Unfortunately we fail to build with an unhelpful error message. MariaDB is now up to version 11.

* Use version.pm to do the version comparison rather than a regex.
* Check for MariaDB and generate a warning if we're going to link against it.
* Correct my email address :-)


Grinnz:
version.pm is for parsing and comparing Perl versions and is not appropriate to use for general program versions (for example, it would interpret a hypothetical '5.10' incorrectly). You could consider versioncmp from Sort::Versions but it's probably not worth adding a dependency for this.

Makefile.PL
```diff
@@ -150,7 +151,18 @@ assert_lib(
   LIBS => $opt->{libs},
 );
 
-print "Looks good.\n\n";
+if ($opt->{libs} =~ /mariadb/) {
+  print <<"MSG";
+
+The chosen MySQL client library appears to be MariaDB's. Compilation may fail.
```
dveeden, line 157:
```suggestion
The chosen MySQL client library appears to be MariaDB's. DBD::mysql doesn't support building against a MariaDB client library.
```

I like this method as it doesn't rely on versions to detect MariaDB (>= 10.0 ) or MySQL (< 10.0) which can be very useful if in the future Oracle would release a MySQL 10.x  (Not unlikely, see the diagram on [this blog post](https://dev.mysql.com/blog-archive/introducing-mysql-innovation-and-long-term-support-lts-versions/) )

dveeden: Not sure what the best text/info is here.

- DBD::mysql requires an Oracle MySQL client library. As the MariaDB client library presents itself as MySQL client library this can be confusing to users. This _might_ get better in the future as `mysql_config` is in `libmariadb-dev-compat` and is optional (at least on Ubuntu 20.04).
- DBD::mysql build with a Oracle MySQL client library can be used to connect to MariaDB.
- Alternatives are DBD::MariaDB and DBD::mysql v4.x. I think only listing the first one (as you did) is probably best, but I'm not sure.
- The MySQL client library must be 8.0 or newer
- Maybe return info on how it was determined to be MariaDB (`mysql_config --libs` matched `mariadb`) to make it easier for people to troubleshoot.

dveeden:
> version.pm is for parsing and comparing Perl versions and is not appropriate to use for general program versions (for example, it would interpret a hypothetical '5.10' incorrectly). You could consider versioncmp from Sort::Versions but it's probably not worth adding a dependency for this.

Thanks, this is very useful info. If MariaDB is detected based on if `mysql_config --libs` output matches `mariadb` then the version matching would be easier as we only need to make sure the first number before the `.` is 8 or higher, we don't have to restrict the upper bound anymore.

As Oracle stayed on 8.0.x for many years and now suddenly has released many new versions like 8.1, 8.2, 8.3, 8.4, 9.0 and 9.1 and is slowly approaching the version range that is used by MariaDB (10.x and up) version checks are becoming more fragile, especially for determining the flavor (MySQL, MariaDB, etc)

dveeden:
Maybe a simple split instead of a regex?
```perl
#!/bin/perl
my @versions = ("3.23.40", "4.1.24", "5.7.40", "8.0.11", "9.1.0", "10.0.0", "11.1.0", "8.0.11-TiDB-v8.4.0");
foreach (@versions) {
	my @verinfo = split(/\./, $_);
	my $verok = @verinfo[0]>=8 ? "ok" : "too old";
	print "$_ has major version @verinfo[0] which is $verok.\n"
}
```

```
3.23.40 has major version 3 which is too old.
4.1.24 has major version 4 which is too old.
5.7.40 has major version 5 which is too old.
8.0.11 has major version 8 which is ok.
9.1.0 has major version 9 which is ok.
10.0.0 has major version 10 which is ok.
11.1.0 has major version 11 which is ok.
8.0.11-TiDB-v8.4.0 has major version 8 which is ok.
```

dveeden (approved):

Makefile.PL
davel: I've kept the text descriptive of the behaviour.

Suggesting Oracle's library seems like good advice. Recommending sticking with 4.x might age badly.

It's not unreasonable for someone to _want_ to use this module with MariaDB. DBD::MariaDB has Unicode fixes that might break your application.

PR #449, opened 2024-10-30
Make ParamValues follow expectations

### LINKED ISSUE #447: segfault using ParamValues from $sth
### DBD::mysql version

4.050

### MySQL client version

8.0.39

### Server version

8.0.39-0ubuntu0.20.04.1

### Operating system version

Ubuntu 20.04.6 LTS

### What happened?

Started work in an older codebase.  I tried using `DBI::Log` module in a test script, and got a segfault.

```
$ perl test.pl
-- Wed Oct 15 13:36:21 2024
--  /home/pyrrhlin/repo/path/DB.pm 160
SET NAMES 'utf8'

Segmentation fault (core dumped)
```

As you can see, the first query, using `$dbh->do()` is fine. The second query is an `$sth->execute(42)` on a `SELECT` statement with one placeholder, an integer id field, that causes the segfault.  (Note: nowhere did we use `$sth->bind_param`.)

I used perl debugger to find the offending statement, and it's [this one](https://metacpan.org/dist/DBI-Log/source/lib/DBI/Log.pm#L192):

```
%values = %{$sth->{ParamValues}};
```

So this module uses the statement handle attribute `ParamValues` to find out what the bound parameter values were.  This is a sometimes supported interface from DBI.  It's [expected to return a hashref](https://metacpan.org/pod/DBI#ParamValues), where
> keys of the hash are the 'names' of the placeholders, typically integers starting at 1. Returns undef if not supported by the driver

However, it should be noted that:

> Some drivers cannot provide valid values for some or all of these attributes until after `$sth->execute` has been successfully called. Typically the attribute will be `undef` in these situations

So, it's clear to me that `DBI::Log` should not be trying to read this data before execution, because for some drivers (including obviously mysql), the data won't be there yet.  Regardless, if the attribute value was `undef`, as DBI docs suggest, the line causing the segfault would not have been executed.  To have entered the `if` block, the attribute must have had some true value.

In the debugger again, I find what it is:
```
  DB<2> c
-- Wed Oct 16 15:17:50 2024
--  /home/pyrrhlin/repo/path/DB.pm 160
SET NAMES 'utf8'

DBI::Log::pre_query(/home/pyrrhlin/repo/path/DBI/Log.pm:197):
197:	        if ($sth && $sth->{ParamValues}) {
  DB<2> print 7 if $sth && $sth->{ParamValues}
7
  DB<3> print ref $sth->{ParamValues}
HASH
  DB<4> x $sth->{ParamValues}
0  HASH(0x559dc82a05f8)
   0 => undef
  DB<5> q
```

So the `ParamValues` attribute had the true value `{ 0 => undef }`.  

However, that data structure was not a "pure-perl" clean hashref, as dereferencing it in list context caused the segfault. Alternative code to copy the hash such as this did not trigger a segfault:
```
my $h = $sth->{ParamValues}; $values{$_} = $h->{$_} for keys %$h`
```

Now, reading [statement handle docs](https://metacpan.org/pod/DBD::mysql#STATEMENT-HANDLES) for `DBD::mysql` there is no mention of `ParamValues` attribute, which raises some question whether it is supported at all.  The docs do confirm however that:

> most attributes are valid only after a successful `execute`. An `undef` value will returned otherwise

Testing shows that after execution, the attribute is populated with e.g., `{ 0 => 42 }`, which is _almost_ useful, but non-compliant with the DBI docs (which required the first argument to be under key `1`). 

So we have these problems:

* under no circumstance should accessing the parameter value, regardless of context, cause a segfault.

* before execution, if we can't provide the values yet, the attribute value should be `undef`.

* after execution, it should probably be using integer keys starting with 1, e.g. `{ 1 => 42 }`

### Other information

Testing performed with system perl on Ubuntu 20.04 LTS :
```
This is perl 5, version 30, subversion 0 (v5.30.0) built for x86_64-linux-gnu-thread-multi
(with 60 registered patches, see perl -V for more detail)
```

Apologies I did not have time to try to replicate the failure with a newer version.  If I get more time I will try to do that.

Issue #353 also suggests trouble with the xs building the data structures tied behind the statement handle attributes.  Perhaps related.

dveeden:
Closes #447 

1. The keys are expected (not guaranteed) to be integers starting at 1. They previously started at 0.
2. The values should be undef if not yet bound

### DIFF
--- a/dbdimp.c
+++ b/dbdimp.c
@@ -4026,9 +4026,16 @@ dbd_st_FETCH_internal(
             I32 keylen;
             for (n= 0; n < DBIc_NUM_PARAMS(imp_sth); n++)
             {
-                keylen= sprintf(key, "%d", n);
-                (void)hv_store(pvhv, key,
-                         keylen, newSVsv(imp_sth->params[n].value), 0);
+                // https://metacpan.org/pod/DBI#ParamValues says keys
+                // are typically integers starting at 1
+                // values should be undef if not yet bound
+                keylen= sprintf(key, "%d", n+1);
+                if (imp_sth->params[n].value) {
+                    (void)hv_store(pvhv, key,
+                             keylen, newSVsv(imp_sth->params[n].value), 0);
+                } else {
+                    (void)hv_store(pvhv, key, keylen, &PL_sv_undef, 0);
+                }
             }
         }
         retsv= sv_2mortal(newRV_noinc((SV*)pvhv));


myrrhlin:
I am not an experienced XS developer, sorry I am perhaps not sufficiently qualified to review this change. 
It looks plausible to handle the keys-starting-at-1 problem. 

The other you mention "The values should be undef if not yet bound" suggests to me a misunderstanding.  My reading of DBI docs are that, before any values are bound or given to `execute`, the `ParamValues` attribute should be `undef`.  It should _not_ be a hashref containing undefined values.  
The only time it should be a hashref with `undef` values, is if someone actually bound `undef` values to the placeholders for `NULL` (either using `bind_param` or `execute` with args).

As for the third problem in issue #447 -- the segfault -- we need a test to reproduce it so that you can find a root cause. I'll try to provide something useful in this regard.

dbdimp.c
```diff
@@ -4026,9 +4026,16 @@ dbd_st_FETCH_internal(
...
+                keylen= sprintf(key, "%d", n+1);
+                if (imp_sth->params[n].value) {
+                    (void)hv_store(pvhv, key,
+                             keylen, newSVsv(imp_sth->params[n].value), 0);
+                } else {
+                    (void)hv_store(pvhv, key, keylen, &PL_sv_undef, 0);
+                }
             }
         }
         retsv= sv_2mortal(newRV_noinc((SV*)pvhv));
```
myrrhlin, line 4041: If I read this correctly, the return value here is always a hashref.  When there are no params saved yet, the return should simply be undef.

dveeden:
> I am not an experienced XS developer, sorry I am perhaps not sufficiently qualified to review this change. It looks plausible to handle the keys-starting-at-1 problem.

Feel free to consider the XS stuff a black box and just verify if the behavior from it matches what you want it to do.

> The other you mention "The values should be undef if not yet bound", and my understanding of the code in this commit, suggests to me there is a misunderstanding. My reading of DBI docs are that, before any values are bound or given to `execute`, the `ParamValues` attribute should be `undef`. It should _not_ be a hashref containing undefined values. The only time it should be a hashref with `undef` values, is if someone actually bound `undef` values to the placeholders for `NULL` (either using `bind_param` or `execute` with args).

"If the driver supports `ParamValues` but no values have been bound yet then the driver should return a hash with placeholders names in the keys but all the values undef, but some drivers may return a ref to an empty hash because they can't pre-determine the names."
(source: https://metacpan.org/pod/DBI#ParamValues)

How I read this, the driver is expected to return a hash with the keys, but values set to `undef` as soon as the statement is prepared.

This seems to match DBD::SQLite:
```perl
#!/bin/perl
use v5.36;
use DBI;
use Data::Dumper;

my $dbh = DBI->connect('dbi:SQLite:dbname=:memory:','','');

my $sth = $dbh->prepare('SELECT ?');
say "after prepare:\n" . Dumper($sth->{ParamValues});
$sth->execute(1);
say "after execute:\n" . Dumper($sth->{ParamValues});

while (my @row = $sth->fetchrow_array()) {
	say "result: " . $row[0];
}
```

```
after prepare:
$VAR1 = {
          '1' => undef
        };

after execute:
$VAR1 = {
          '1' => 1
        };

result: 1
```

> As for the third problem in issue #447 -- the segfault -- we need a test to reproduce it so that you can find a root cause. I'll try to provide something useful in this regard.

Thanks

myrrhlin:
You are correct, I was mistaken about the expected behavior.  I was misled by
> Some drivers cannot provide valid values for some or all of these attributes until after $sth->execute has been successfully called. Typically the attribute will be undef in these situations

And `DBD::mysql` has similar language:
> Note, that most attributes are valid only after a successful execute. An undef value will returned otherwise

but that is the default expectation for all attributes, and _this_ one is specially described!.

Sorry for my confusion. Thanks for your patience. I will update my tests.

PR #457, opened 2025-01-03
Use safer quoting for placeholders

mattlaw:
Switch to mysql_real_escape_string_quote for placeholder replacement, allowing placeholders to be used when NO_BACKSLASH_ESCAPES is in effect.

### DIFF
--- a/dbdimp.c
+++ b/dbdimp.c
@@ -636,7 +636,7 @@ static char *parse_params(
             if (!is_num)
             {
               *ptr++ = '\'';
-              ptr += mysql_real_escape_string(sock, ptr, valbuf, vallen);
+              ptr += mysql_real_escape_string_quote(sock, ptr, valbuf, vallen, '\'');
               *ptr++ = '\'';
             }
             else
--- a/t/17quote.t
+++ b/t/17quote.t
@@ -23,13 +23,19 @@ my @results_ansi = (qw/ 'foo' 'foo\'bar' 'foo\\\\bar'/);
 my @results_no_backlslash = (qw/ 'foo' 'foo''bar' 'foo\\bar'/);
 my @results = (\@results_empty, \@results_ansi, \@results_no_backlslash);
 
-plan tests => (@sqlmodes * @words * 2 + 1);
+plan tests => (@sqlmodes * @words * 3 + 1);
 
 while (my ($i, $sqlmode) = each @sqlmodes) {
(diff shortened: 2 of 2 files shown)


dbdimp.c
```diff
@@ -636,7 +636,7 @@ static char *parse_params(
             if (!is_num)
             {
               *ptr++ = '\'';
-              ptr += mysql_real_escape_string(sock, ptr, valbuf, vallen);
+              ptr += mysql_real_escape_string_quote(sock, ptr, valbuf, vallen, '\'');
```
dveeden, line 639: So here the quoting context is set to `'`. Are we sure this is always correct? Otherwise `mysql_real_escape_string()` could have been implemented with a call to `mysql_real_escape_string_quote()` with a constant as quoting context.

dveeden:
What about the about `ANSI_QUOTES` note on https://dev.mysql.com/doc/c-api/9.1/en/mysql-real-escape-string-quote.html  ?

dveeden:
Please add any info on the usecase for this and why it would be safer etc.

So for this looks like an ok change to me.

dbdimp.c
mattlaw: Yes, the quote in use is hardcoded on the previous line. This doesn't generalise to _every_ possible caller, so it's not safe to generalise `mysql_real_escape_string` itself.

mattlaw:
> Please add any info on the usecase for this and why it would be safer etc.
> 
If you run t/17quote.t on the previous version of the code, you can see that NO_BACKSLASH_ESCAPES causes placeholder replacement to produce invalid query syntax (effectively the `?` placeholder is replaced by a single `'` because `mysql_real_escape_string` returns -1 and rewinds the pointer).

```
ok 20 - NO_BACKSLASH_ESCAPES foo
DBD::mysql::db selectrow_array failed: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ''' at line 1 at t/17quote.t line 34.
DBD::mysql::db selectrow_array failed: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ''' at line 1 at t/17quote.t line 34.
# Looks like your test exited with 255 just after 20.
```

mattlaw:
> What about the about `ANSI_QUOTES` note on https://dev.mysql.com/doc/c-api/9.1/en/mysql-real-escape-string-quote.html ? Could this cause a regression for `ANSI_QUOTES` with double quoted characters?

In this context, we're never processing a double-quoted string. We've added single quotes ourselves for quoting purposes.

PR #465, opened 2025-02-09
Ensure MySQL is only initialized once

roavin:


Grinnz:
Pardon my ignorance on the issue, but will this work on perls build without pthreads? (On recent perls, this now has to be explicitly requested, but it was the default on very old perls)

Grinnz:
Here's an example of a check for whether a Perl has been compiled with pthreads: https://metacpan.org/release/OLEG/Net-DNS-Native-0.22/source/Makefile.PL

I do not suggest requiring pthreads as this module does, because it's not central to DBD::mysql's operation. But perhaps this code needs to be conditional on it somehow

roavin:
It's certainly possible to do this without pthreads; perhaps the perl runtime or MySQL client library ship with synchronization primitives that could be used but I personally am not familiar enough with them to know if that is the case or not. I used pthreads here because it provided the necessary facility and is, as far as I'm aware, fairly universally available.

roavin:
> Maybe using [`BOOT`](https://perldoc.perl.org/perlxs#The-BOOT:-Keyword) would be sufficient?

Probably not - since each MPM worker thread spawns its own Perl interpreter, `BOOT` would possibly be executed simultaneously as well (actually, it might make the bug quicker to trigger since it no longer depends on requests that connect to the DB, which could be a benefit in its own way 🙃 )

It looks like neither Perl's XS interface nor libmysql export any synchronization primitive that can be used here, which is unfortunate. 

How likely is the situation that somebody uses DBD::mysql with a Perl or system that doesn't have pthreads? Would it be reasonable to add the check (which makes the build fail without pthreads) as suggested above by @Grinnz ?

Grinnz:
It is quite likely on Perls older than 5.20 that DBD::mysql would be used on a Perl without pthreads linked. I don't think it's a good idea to break this configuration.

Grinnz:
I also am not familiar with how pthreads works on Windows, as you can see in the module I linked there is some special casing around that.

roavin:
> It is quite likely on Perls older than 5.20 that DBD::mysql would be used on a Perl without pthreads linked. I don't think it's a good idea to break this configuration.

5.20 is over 10 years old now, though. How likely is it to have such an old Perl version with such a new version of DBD::mysql?

> I also am not familiar with how pthreads works on Windows, as you can see in the module I linked there is some special casing around that.

pthreads isn't shipped with Windows directly, unlike on Linux where pthreads is pretty much always available when a compiler is available. The module you linked includes the prebuilt lib files for Windows, which is why it has the special case. For something like pthreads, that's probably the simplest working approach that doesn't introduce other additional dependencies.

dveeden:
For now it looks like:
- Some `ifdef` magic might be needed to make this work on installations without threading
- Check if there is anything in Perl that abstracts threads etc.

This seems somewhat similar: https://stackoverflow.com/a/39136689/958681

Note that DBD::mysql on Windows is currently unfortunately mostly not working due to how Perl, MySQL and OpenSSL are build and shipped. However we shouldn't make that even more difficult if we can avoid it.

roavin:
The Win32 path I described above would use the [`InitOnceExecuteOnce`](https://learn.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-initonceexecuteonce) API, which pretty much does the same thing as `pthread_once`. That function is available starting with Windows Vista or Windows Server 2008, which are both long out of support (2017 and 2020, respectively). I presume that this won't be an issue, particularly if Windows support is wonky anyway.

I do need to set up a build environment for this, though; Windows programming is my "home territory" for work, but not for the Perl work I do in my free time. I assume it won't be particularly difficult.

dveeden:
@roavin Please see [this comment](https://github.com/perl5-dbi/DBD-mysql/pull/384#issuecomment-2548585132) and the rest of that thread for the current state on Windows.

Makefile.PL
```diff
@@ -198,6 +198,16 @@ if ($^O eq 'VMS') {
 }
 $cflags .= " -DDBD_MYSQL_NO_CLIENT_FOUND_ROWS" if $opt->{'nofoundrows'};
 $cflags .= " -g ";
+
+unless ($^O eq 'MSWin32') {
+  if ($Config{usethreads} || $Config{osname} =~ /darwin|linux/i || $Config{libs} =~ /-l?pthread\b/ || $Config{ldflags} =~ /-l?pthread\b/) {
```
Grinnz, line 203: I'm not sure this condition is correct, as it will assume that any mac or linux install has pthreads which is not true?

Grinnz: Though I see that is also done in the condition I linked to, seems strange.

roavin: Yes, that's why I did it that way. I can imagine that OSX systems generally have pthreads and I suppose the thinking for Linux is that pretty much any system not older than 2 decades will have pthreads there too.

maxbarry:
Just posting to say this would be really helpful; the underlying bug locked up our site today.

dveeden:
@roavin could you fix the conflict? Maybe we can move this forward again and get this merged at some point.

@maxbarry could you give some more details about what happened exactly and with what versions?

maxbarry:
Ubuntu 24.04.4 LTS
DBD::mysql 4.052
mysql client 8.0.46-0ubuntu0.24.04.4 for Linux on x86_64 ((Ubuntu))
mysql server 8.0.46-0ubuntu0.22.04.3 

Periodically an Apache process pegs at maximum CPU load (500% on our machine). Inspection via the `server-status` Apache mod reveals all its threads are marked "G" - gracefully finishing. It stays like this forever unless killed with `-9`. Happens apparently randomly days / weeks / months apart. Usually we have a script to detect and forcibly kill the offending process. Has been happening for at least a year, possibly much longer.

PR #467, opened 2025-03-18
Add tests for UTF8 characters in table definition and errors 

michal-josef-spacek:
Add tests for UTF8 characters in table definition and errors
The tests are taken from DBD::MariaDB and adopted to DBD::mysql.

Testing is provided on Japanese and Chinese characters.

There are two test files:
* one for testing of utf8/utf8mb4 identifiers
* one for testing of utf8 error messages

There are three runs of all tests in both test files:
* one without mysql_enable_utf8 option
* one with enabled mysql_enable_utf8 option
* one with enabled mysql_enable_utf8mb4 option

There are XXX comments for situations which not follow
mysql_enable_utf8/mb4 option.

### DIFF
--- a/MANIFEST
+++ b/MANIFEST
@@ -53,6 +53,8 @@ t/51bind_type_guessing.t
 t/52comment.t
 t/53comment.t
 t/55utf8.t
+t/55utf8_errors.t
+t/55utf8_identifiers.t
 t/55utf8mb4.t
 t/56connattr.t
 t/57trackgtid.t
--- a/t/55utf8_errors.t
+++ b/t/55utf8_errors.t
@@ -1,0 +1,111 @@
+use strict;
+use warnings;
+
+use Test::More;
+use DBI;
+use Encode;
+
+use vars qw($test_dsn $test_user $test_password);
+use lib 't', '.';
+require "lib.pl";
+
+sub skip_rt_102404 {
+    skip "(Perl 5.13.1 and DBI 1.635) or DBI 1.639 is required due to bug RT 102404", $_[0] unless ($] >= 5.013001 and eval { DBI->VERSION(1.635) }) or eval { DBI->VERSION(1.639) };
+}
+
+my $dbh;
+eval {
+    $dbh = DBI->connect($test_dsn, $test_user, $test_password,
+                      { RaiseError => 1, PrintError => 1, AutoCommit => 0 });
+};
+if ($@) {
+    plan skip_all => "no database connection";
+}
+$dbh->disconnect();
+
+plan tests => 10 * 3;
+
+# All in internal Perl Unicode
+my $jpnErr = qr/\x{4ed8}\x{8fd1}.*\x{884c}\x{76ee}/; # Use \x{...} instead \N{U+...} due to Perl 5.12.0 bug
+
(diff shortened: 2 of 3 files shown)


michal-josef-spacek:
PR is finished, any feedback?

dveeden (approved):
This looks ok to me on first glance. Is there a bigger plan or are you only adding tests?

dveeden:
@michal-josef-spacek 

-  I assume you have already found/read https://blogs.perl.org/users/grinnz/2023/12/migrating-from-dbdmysql-to-dbdmariadb.html ?
- I'm all for updating tests and documentation
-  https://github.com/perl5-dbi/DBD-mysql/issues?q=is%3Aissue%20state%3Aopen%20label%3Autf8 lists all UTF-8 related issues.
- My understanding about the big revert is that the changes made in 4.041 were breaking existing scripts for too many people and then basically the whole release was reverted including things that might have been unrelated. This then lead to the DBD::MariaDB / DBD::mysql split etc.

Things to consider when working on UTF-8:
- Compatibility with existing scripts etc is important. This might need a feature flag and a reasonable upgrade path for users.
- In MySQL there is `utf8mb3` and `utf8mb4`. And there is `utf8` which is *currently* a alias for `utf8mb3`. This is confusing as outside of MySQL nobody really uses 3-byte UTF-8. The `utf8` alias might change at some point to point to `utf8mb4` and `utf8mb3` might be removed at some point. Where possible I would like to have a single flag to enable `UTF-8` in DBD::mysql instead of multiple.

dveeden:
cc @Grinnz

michal-josef-spacek:
@dveeden 

> * I assume you have already found/read https://blogs.perl.org/users/grinnz/2023/12/migrating-from-dbdmysql-to-dbdmariadb.html ?

I read this. It's misleading from my side, too. The usage of utf8::upgrade and utf8::downgrade is a very problematic thing at all.

> * I'm all for updating tests and documentation

yes, I am for. Working on it.

> * https://github.com/perl5-dbi/DBD-mysql/issues?q=is%3Aissue%20state%3Aopen%20label%3Autf8 lists all UTF-8 related issues.

I know.

> * My understanding about the big revert is that the changes made in 4.041 were breaking existing scripts for too many people and then basically the whole release was reverted including things that might have been unrelated. This then lead to the DBD::MariaDB / DBD::mysql split etc.

Yes, I know. Too many changes at one release are the biggest issue.
Still, I don't see a report that defines the issue (still in progress from my side). My actual understanding is the usage of bad techniques to create utf8 characters. sic!

> Things to consider when working on UTF-8:
> * Compatibility with existing scripts etc is important. This might need a feature flag and a reasonable upgrade path for users.

Yes, I know.

> * In MySQL there is `utf8mb3` and `utf8mb4`. And there is `utf8` which is _currently_ a alias for `utf8mb3`. This is confusing as outside of MySQL nobody really uses 3-byte UTF-8. The `utf8` alias might change at some point to point to `utf8mb4` and `utf8mb3` might be removed at some point. Where possible I would like to have a single flag to enable `UTF-8` in DBD::mysql instead of multiple.

Yes, I know. This is a bit confusing, but I think this is possible. 
There are bigger issues, in code and doc too.

Issue #468, opened 2025-04-09
The mentioning of MariaDB in the documentation


michal-josef-spacek:
Hi, I would like to discuss information about MariaDB in DBD::mysql documentation.

My intention with this is to understand what we support and what is buggy or missing.

**There are some facts**:
* There is no information about MariaDB in DBD::mysql documentation
* Tests in CI are running against MySQL only
* Tests are written against MySQL only (instead of 57trackgtid.t. which has ```$dbh->{mysql_serverversion} > 100000``` condition, which is a bit tricky (The version of MySQL could be greater than 100000 in the future).
* I have a situation where I have DBD::mysql compiled with libmysqlclient.so.21 on Fedora 41, connected to MariaDB, and result of the ```$dbh->get_info($GetInfoType{SQL_DBMS_NAME})``` is ```MySQL``` with version ```$dbh->{mysql_serverversion}``` is ```50505``` (With DBD::MariaDB and same libmysqlclient.so.21 library I have "MariaDB" and "101111").
* In the real world in Red Hat RHEL: In RHEL8 we have a perl-DBD-mysql only and MySQL and MariaDB databases. So there were only one option to connect to MariaDB and this was DBD::mysql. From RHEL9 we have perl-DBD-MariaDB, so situation is "better". 

**My questions**:
* Do we support working with MariaDB database?
* Or specify the situation. e.g. We support work with MariaDB to/from version etc.

**My proposals**:
* Add some information about MariaDB to the DBD::mysql documentation. It depends on what we write there, according to the answers above.

dveeden:
When compiling DBD::mysql it requires MySQL v8.0 or newer (except for v4.x versions of DBD::mysql). Building against MariaDB isn't supported and doesn't work as the C API of MySQL and MariaDB have diverted.

But once DBD::mysql is build it should be able to connect to MariaDB and function. The same is true for connecting with DBD::mysql to TiDB or anything else that supports the MySQL protocol and syntax.

Limitations here are that very old versions of MySQL (and MariaDB) might not work. (e.g. MySQL 3.23.x, 4.0, etc). I would expect the v10 protocol and `CLIENT_PROTOCOL_41`.

michal-josef-spacek:
@dveeden Ok, thank you for the answer. I understand it. I am proposing to add this information to the documentation because this is not clean. **Is this ok for you?** I would prepare doc change.

[...]

PR #469, opened 2025-04-10
Correct SQL_DRIVER_NAME

dveeden:
Ref #468


dveeden:
@Tux do you have the permission to remove/disable appveyor?

PR #476, opened 2025-05-09
#475: Improve error message for unsupported named parameters

### LINKED ISSUE #475: Improve error message for (unsupported) named parameter usage
Lyrion Music Server added some new database queries using named parameters in the statement.  This is supported by SQLite (via DBD-SQLite) but this is not available to DBD-mysql/MySQL (or MariaDB).  When the query is executed the reported error is:

```
[25-05-02 18:47:27.2333] Slim::Schema::Storage::throw_exception (122) Error: DBI Exception: DBD::mysql::st bind_param failed: Illegal parameter number [for Statement "
                                        SELECT contributor_album.role AS role, contributors.name AS name, contributors.id AS id
                                        FROM contributor_album
                                        JOIN contributors ON contributors.id = contributor_album.contributor
                                        WHERE contributor_album.album = :album
                                        AND contributor_album.role IN (5,1)
                                        ORDER BY contributor_album.role, contributors.namesort
                                "]
[25-05-02 18:47:27.2343] Slim::Schema::Storage::throw_exception (122) Backtrace:
```

Full report: https://github.com/LMS-Community/slimserver/issues/1265

It took a bit of digging to discover this.  It seems possible to add a more helpful error message which turns the error in to:

```
    [25-05-06 20:46:04.2629] Slim::Schema::Storage::throw_exception (122) Error: DBI Exception: DBD::mysql::st bind_param failed: named parameters are unsupported: :album [for Statement "
                                            SELECT contributor_album.role AS role, contributors.name AS name, contributors.id AS id
                                            FROM contributor_album
                                            JOIN contributors ON contributors.id = contributor_album.contributor
                                            WHERE contributor_album.album = :album
                                            AND contributor_album.role IN (5,1)
                                            ORDER BY contributor_album.role, contributors.namesort
                                    "]
```

The change is based on the named parameter handling in DBD-SQLite: https://github.com/DBD-SQLite/DBD-SQLite/blob/60515b30f8b816cb09e9f40f2f0a5829af031899/dbdimp.c#L1559

JKDingwall:
Previously just 'Illegal parameter number' was logged, make this a more helpful message.

```
[25-05-06 20:46:04.2629] Slim::Schema::Storage::throw_exception (122) Error: DBI Exception: DBD::mysql::st bind_param failed: named parameters are unsupported: :album [for Statement "
                                        SELECT contributor_album.role AS role, contributors.name AS name, contributors.id AS id
                                        FROM contributor_album
                                        JOIN contributors ON contributors.id = contributor_album.contributor
                                        WHERE contributor_album.album = :album
                                        AND contributor_album.role IN (5,1)
                                        ORDER BY contributor_album.role, contributors.namesort
                                "]
```

### DIFF
--- a/dbdimp.c
+++ b/dbdimp.c
@@ -4211,6 +4211,23 @@ int dbd_bind_ph(SV *sth, imp_sth_t *imp_sth, SV *param, SV *value,
 
   if (param_num <= 0  ||  param_num > DBIc_NUM_PARAMS(imp_sth))
   {
+    if (!looks_like_number(param))
+    {
+      STRLEN len;
+      char *paramstring;
+      paramstring = SvPV(param, len);
+      if(paramstring[len] == 0 && strlen(paramstring) == len)
+      {
+        do_error(sth, JW_ERR_ILLEGAL_PARAM_NUM, form("named parameters are unsupported: %s", paramstring), NULL);
+        return FALSE;
+      }
+      else
+      {
+        do_error(sth, JW_ERR_ILLEGAL_PARAM_NUM, "<param> could not be coerced to a C string", NULL);
+        return FALSE;
+      }
+    }
+
     do_error(sth, JW_ERR_ILLEGAL_PARAM_NUM, "Illegal parameter number", NULL);
     return FALSE;
   }
--- a/lib/DBD/mysql.pm
+++ b/lib/DBD/mysql.pm
@@ -1725,6 +1725,11 @@ parameter value and C<execute> the statement again, with other values
 unchanged. The attribute remains properly populated after the C<finish> 
 method is called, with the values from the last execution.
 
+MySQL does not support named place holders in C<bind_param>.  If a
+string is passed to C<bind_param> as the parameter index then a
+`JW_ERR_ILLEGAL_PARAM_NUM` with a message of "named parameters are
+unsupported" error is reported.
+
 =item mysql_gtids
 
 Returns GTID(s) if GTID session tracking is ensabled in the server via
--- a/t/45bindnamedparam_error.t
+++ b/t/45bindnamedparam_error.t
@@ -1,0 +1,68 @@
+use strict;
+use warnings;
+
+use Test::More;
+use DBI;
+use vars qw($test_dsn $test_user $test_password);
+use lib 't', '.';
+require 'lib.pl';
+
+my $dbh;
+eval {$dbh = DBI->connect($test_dsn, $test_user, $test_password,
+  { RaiseError => 1, AutoCommit => 1}) or ServerError();};
+
+if ($@) {
+    plan skip_all => "no database connection";
+}
+plan tests => 15;
+
+SKIP: {
+    ok $dbh->do('SET @@auto_increment_offset = 1');
+    ok $dbh->do('SET @@auto_increment_increment = 1');
+}
+
+my $create= <<EOT;
+CREATE TEMPORARY TABLE dbd_mysql_t45bindnamedparam (
+    id INT NOT NULL AUTO_INCREMENT PRIMARY KEY,
+    num INT(3))
+EOT
+
+ok $dbh->do($create), "create table dbd_mysql_t45bindnamedparam";
+
+ok $dbh->do("INSERT INTO dbd_mysql_t45bindnamedparam VALUES(NULL, 1)"), "insert into dbd_mysql_t45bindnamedparam (null, 1)";
+
+my $rows;
+ok ($rows= $dbh->selectall_arrayref("SELECT * FROM dbd_mysql_t45bindnamedparam"));
+
+is $rows->[0][1], 1, "\$rows->[0][1] == 1";
+
+my $sth;
+ok ($sth = $dbh->prepare("SELECT * FROM dbd_mysql_t45bindnamedparam WHERE num = :num"));
+
+$dbh->{PrintError} = 0;
+$dbh->{PrintWarn} = 0;
+eval {($sth->bind_param(":num", 1, SQL_INTEGER()));};
+$dbh->{PrintError} = 1;
+$dbh->{PrintWarn} = 1;
+ok defined($DBI::errstr);
+
+like($DBI::errstr, qr/named parameters are unsupported/, 'bind_param reports expected error with named parameter (string type)');
+
+ok ($sth->finish());
+
+
+ok ($sth = $dbh->prepare("SELECT * FROM dbd_mysql_t45bindnamedparam WHERE num = :num"));
+
+$dbh->{PrintError} = 0;
+$dbh->{PrintWarn} = 0;
+eval {($sth->bind_param("\0:num", 1, SQL_INTEGER()));};
+$dbh->{PrintError} = 1;
+$dbh->{PrintWarn} = 1;
+ok defined($DBI::errstr);
+
+like($DBI::errstr, qr/could not be coerced to a C string/, 'bind_param reports expected error with named parameter (non-string type)');
+
+ok ($sth->finish());
+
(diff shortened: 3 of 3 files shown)


dveeden:
This probably need a few doc/pod updates and one or more tests.

dveeden:
@JKDingwall 

1. Yes just check the current docs and update if needed. Preferably with some suggestion about what to use instead. If you think no updates are needed than that might be ok.
2. Yes. Probably a new test file to test that this does work as intended. If there is a file that already has something very similar that including this in that file would also be fine.

Also note that MySQL might make it easier to add named attributes in the future:
- https://dev.mysql.com/doc/refman/9.3/en/query-attributes.html
- https://dev.mysql.com/doc/c-api/9.3/en/mysql-bind-param.html

JKDingwall:
@dveeden I've added a new test that the expected error message is reported if `bind_param(':name', 1)` is called.  I've also added a couple of sentences to the documentation that named parameters are not supported.  Please let me know you'd like any changes made, thanks!

This is the output from the test:

```
t/45bindnamedparam_error.t ....................
1..11
ok 1
ok 2
ok 3 - create table dbd_mysql_t45bindnamedparam
ok 4 - insert into dbd_mysql_t45bindnamedparam (null, 1)
ok 5
ok 6 - $rows->[0][1] == 1
ok 7
ok 8
ok 9 -
ok 10
ok 11
Argument ":num" isn't numeric in subroutine entry at t/45bindnamedparam.t line 50.
DBD::mysql::st bind_param failed: named parameters are unsupported: :num at t/45bindnamedparam.t line 50.
ok
```

I had seen the documentation about named attributes but after some research I realised that its actually a bit misleading.  The feature provides a way to annotate a query with, for example, the name of script which invoked the query.

- https://stackoverflow.com/questions/77626696/understanding-mysql-stmt-bind-named-param-mysql-c-api
- https://mysql.wisborg.dk/2021/02/02/mysql-query-attributes/

t/45bindnamedparam_error.t
```diff
@@ -0,0 +1,53 @@
...
+eval {$dbh = DBI->connect($test_dsn, $test_user, $test_password,
+  { RaiseError => 1, AutoCommit => 1}) or ServerError();};
+
+if ($@) {
+    plan skip_all => "no database connection";
+}
+plan tests => 11;
+
+SKIP: {
+    skip 'SET @@auto_increment_offset needs MySQL >= 5.0.2', 2 unless $dbh->{mysql_serverversion} >= 50002;
```
dveeden, line 20: DBD::mysql needs MySQL 8.0 or newer, so this shouldn't be needed. But on the other hand... it doesn't hurt either.

dbdimp.c
```diff
@@ -4209,6 +4209,23 @@ int dbd_bind_ph(SV *sth, imp_sth_t *imp_sth, SV *param, SV *value,
...
+      char *paramstring;
+      paramstring = SvPV(param, len);
+      if(paramstring[len] == 0 && strlen(paramstring) == len)
+      {
+        do_error(sth, JW_ERR_ILLEGAL_PARAM_NUM, form("named parameters are unsupported: %s", paramstring), NULL);
+        return FALSE;
+      }
+      else
+      {
+        do_error(sth, JW_ERR_ILLEGAL_PARAM_NUM, "<param> could not be coerced to a C string", NULL);
```
dveeden, line 4224: Is there a test for this error?

What if there are multiple params. Does it indicate which one is the issue? Could/should it?

lib/DBD/mysql.pm
```diff
@@ -1725,6 +1725,10 @@ parameter value and C<execute> the statement again, with other values
 unchanged. The attribute remains properly populated after the C<finish> 
 method is called, with the values from the last execution.
 
+MySQL does not support named place holders in C<bind_param>.  If a
+string is passed to C<bind_param> as the parameter index then a
+"named parameters are unsupported" error is reported.
```
dveeden, line 1730:
```suggestion
`JW_ERR_ILLEGAL_PARAM_NUM` with a message of "named parameters are unsupported" error is reported.
```

btw. I think `JW` stands for Jim Winstead. See `git log`

dveeden (approved):

lib/DBD/mysql.pm
JKDingwall: Thank you for the suggestion, applied.

t/45bindnamedparam_error.t
JKDingwall: I copied this from t/40bindparam2.t which I used as a template but I have removed it.

dbdimp.c
JKDingwall: I have added a test for the `else` condition.

From the test execution:
```
Argument ":num" isn't numeric in subroutine entry at t/45bindnamedparam_error.t line 44.
DBD::mysql::st bind_param failed: named parameters are unsupported: :num at t/45bindnamedparam_error.t line 44.
Argument "\0:num" isn't numeric in subroutine entry at t/45bindnamedparam_error.t line 58.
DBD::mysql::st bind_param failed: <param> could not be coerced to a C string at t/45bindnamedparam_error.t line 58.
```

The name of parameter `:num` is reported twice for the  first test, and once for the second.  I'm not sure if the "isn't numeric in subroutine entry" message comes from `SvIV()` or `looks_like_number()`.  I didn't want to try printing the string for the second case in the DBD-mysql error message in case it had weird escape characters.

dveeden: Thanks for the explanation and the second test.

The `Argument "..." isn't numeric in subroutine entry` warning comes from `SvIV(param)` at `dbdimp.c:4190`, which runs before the `looks_like_number` check — `looks_like_number` itself is silent. The `"in subroutine entry"` suffix is Perl's standard label when the warning fires during an XSUB call. That's why `:num` appears twice in the first test (once from SvIV, once from our `do_error`) and once in the second (SvIV warns, but our `else` branch deliberately omits the value).

Your reasoning for not printing the string in the `else` branch is right — once `paramstring[len] != 0` or `strlen != len`, feeding it to `form("%s", ...)` would either truncate at the embedded NUL or read past the buffer. If you want to surface something useful there, `form("named parameters are unsupported: %.*s", (int)len, paramstring)` is bounded by `len` and safe even with embedded NULs (though non-printables would still come through raw).

If the doubled `:num` in the first test bothers you, you can hoist the `looks_like_number(param)` check above the `SvIV(param)` at line 4190 and short-circuit — then SvIV never runs for the named-parameter case and only our error is emitted. Happy to take it either way; the current behavior is fine with me.

*(This reply was drafted with AI assistance.)*

dveeden: I'm ok with merging this as it is now. Or do you want to do one more iteration on it?

JKDingwall: > Your reasoning for not printing the string in the else branch is right — once paramstring[len] != 0 or strlen != len, feeding it to form("%s", ...) would either truncate at the embedded NUL or read past the buffer. If you want to surface something useful there, form("named parameters are unsupported: %.*s", (int)len, paramstring) is bounded by len and safe even with embedded NULs (though non-printables would still come through raw).

If the `else` case is reached I'd imagine it would be due to a programming error other than attempting to use a `:named` parameter.  If that is the case I'm not sure that printing `paramstring` would provide any useful output.

> If the doubled :num in the first test bothers you, you can hoist the looks_like_number(param) check above the SvIV(param) at line 4190 and short-circuit — then SvIV never runs for the named-parameter case and only our error is emitted. Happy to take it either way; the current behavior is fine with me.

I think this would make the code less tidy for the sake of a (not even) minor cosmetic change to the  message.

If you're happy then I'm happy:)

